PR #29, opened 2019-07-27
run vagrant with admin rights to

pragmaticmind:
Related to this ticket: https://github.com/holochain/holonix/issues/18#issuecomment-515690984
npm wants to add symlinks to the shared windows drive in virtualbox
this is only possible with an adapted vagrant file
and you have to run vagrant up (virtualbox) as administrator


pragmaticmind:
option rejected by Davidmeister because of the downside of running as admin. Preferable use nix scripts to work around this problem

PR #34, opened 2019-08-30
remove nix-env

freesig:
This removes `nix-env` from the quick start altogether. It was causing some confusion and @yegortimoshenko has pointed out some issue that could come from using it.
The only side effect of this is that anyone using an IDE will need to launch it from within nix-shell.
I have added a brief section about that but perhaps it should be more detailed and list how to launch all the common editors?

### DIFF
--- a/nix.html
+++ b/nix.html
@@ -93,7 +93,7 @@
                                 </ul>
                                 <p>It is important that these remain consistent across compatible apps and the Holochain Core.</p>
                                 <p>The Holonix repository tracks standard, shared dependencies for all of the scenarios in which we use NixOS. Typically you won&rsquo;t need to interact with Holonix directly.</p>
-                                <p>The main Nix tool used in Holochain development workflows is <span class="inline-code">nix-shell</span>, a managed Bash shell that overlays a new environment and set of tools on top of your existing environment</p>
+                                <p>The main Nix tool used in Holochain development workflows is <span class="inline-code">nix-shell</span>, a managed Bash shell that overlays a new environment and set of tools on top of your existing environment.</p>
                                 <p>Many popular package management tools only target a single OS. NixOS package management supports most OSes.</p>
                                 <p>The full suite of Nix tooling is broad and deep. There&rsquo;s even a dedicated OS and functional programming language. Learn more with the <a href="https://nixos.wiki/wiki/Main_Page" target="_blank" rel="noopener">NixOS Wiki</a> or the <a href="https://nixos.org/nixos/nix-pills/" target="_blank" rel="noopener">Pills</a> Tutorial. The community IRC chat at <span class="inline-code">#nixos</span> on freenode is active and helpful.</p>
                                 <br><h2 class="section-title" id="nix-shell"><span class="inline-code">nix-shell</span></h2><br>
@@ -102,12 +102,6 @@
                                 <p>You can extend the master <span class="inline-code">default.nix</span> file or use it directly from <a href="https://holochain.love/" target="_blank" rel="noopener">https://holochain.love</a>. Opening a shell from holochain.love downloads updates automatically. The initial run will take some time to download and build. It gets much faster on subsequent runs.</p>
                                 <p><span class="inline-code">nix-shell</span> cleans up everything it added when you exit. You need to re-enter the shell each time you want to work.</p>
                                   <p><em>Note: The files are cached locally on your machine, so they will not be re-downloaded or rebuilt the next time you enter the shell.</em></p>
-                                <h4 id="Troubleshooting">Troubleshooting</h4>
-                                <p>By default, <span class="inline-code">nix-shell</span> only adds new things to your bash shell. Your existing setup could contain dependencies that override <span class="inline-code">nix-shell</span> and conflict with expectations.</p>
-                                <p>The following command tells <span class="inline-code">nix-shell</span> to attempt to isolate common conflicts.</p>
-                                <p><code>nix-shell --pure</code></p>
-                                  <p><em>Try this option first if you&rsquo;re seeing errors where others are not.</em></p>
-                                <p>The &ldquo;nix pod&rdquo; in Holochain core isolates more aggressively for debugging.</p>
                             </div>
 
                         </section>
--- a/nix.html
+++ b/nix.html
@@ -44,12 +44,10 @@
     .copy-confirmation {
         margin-left: 10px;
     }
-
     .code-long .copy-to-clipboard { margin: 5px 0 0 0; display: inline-block; }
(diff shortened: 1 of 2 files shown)


nix.html
```diff
@@ -88,35 +86,28 @@ <h1 class="doc-title">More Info on NixOS</h1>
...
+                                <p><span class="inline-code">nix-shell</span> cleans up everything it added when you exit. You need to re-enter the shell each time you want to work.</p>
                                   <p><em>Note: The files are cached locally on your machine, so they will not be re-downloaded or rebuilt the next time you enter the shell.</em></p>
                                 <h4 id="Troubleshooting">Troubleshooting</h4>
-                                <p>By default, <span class="inline-code">nix-shell</span> only adds new things to your bash shell. Your existing setup could contain dependencies that override <span class="inline-code">nix-shell</span> and conflict with expectations.</p>
-                                <p>The following command tells <span class="inline-code">nix-shell</span> to attempt to isolate common conflicts.</p>
+                                <p>By default, <span class="inline-code">nix-shell</span> only adds new things to your bash shell. Your existing setup could contain dependencies that override <span class="inline-code">nix-shell</span> and conflict with expectations.</p>
```
yegortimoshenko: `nix-shell` dependencies will always be prioritized if `nix-shell` is correctly installed. Instead of running `nix-shell --pure` (which is very impractical, because it removes developer tools such as editor from `PATH`), I think we should explain the user how to fix their environment instead.

nix.html
```diff
@@ -88,35 +86,28 @@ <h1 class="doc-title">More Info on NixOS</h1>
...
                                 </ul>
                                 <p>It is important that these remain consistent across compatible apps and the Holochain Core.</p>
                                 <p>The Holonix repository tracks standard, shared dependencies for all of the scenarios in which we use NixOS. Typically you won&rsquo;t need to interact with Holonix directly.</p>
-                                <p>The main Nix tools used in Holochain development workflows are:</p>
-                                <ul>
-                                  <li><span class="inline-code">nix-shell</span>, a managed Bash shell that overlays a new environment and set of tools on top of your existing environment</li>
-                                  <li><span class="inline-code">nix-env</span>, a cross-platform package manager that installs the tools right into your existing environment</li>
-                                </ul>
+                                <p>The main Nix tool used in Holochain development workflows is <span class="inline-code">nix-shell</span>, a managed Bash shell that overlays a new environment and set of tools on top of your existing environment</p>
```
yegortimoshenko:
```suggestion
                                <p>The main Nix tool used in Holochain development workflows is <span class="inline-code">nix-shell</span>, a managed Bash shell that overlays a new environment and set of tools on top of your existing environment.</p>
```

nix.html
```diff
@@ -88,35 +86,28 @@ <h1 class="doc-title">More Info on NixOS</h1>
...
-                                <p>You can extend the master <span class="inline-code">default.nix</span> file or use it directly from <a href="https://holochain.love/" target="_blank" rel="noopener">https://holochain.love</a>. Opening a shell from holochain.love downloads updates automatically. The initial run will take some time to download and build. It gets much faster on subsequent runs.</p>
-                                <p><span class="inline-code">nix-shell</span> cleans up everything it added when you exit. You need to re-enter the shell each time you want to work.</p>
+                                <p>While working on Holochain, you will usually have an active <span class="inline-code">nix-shell</span> to run commands. All the extra dependencies and environment variables will be cleaned up automatically when you close the shell.</p>
+                                <p><span class="inline-code">nix-shell</span> will give you the latest releases every time you enter it. Nix is configured by <span class="inline-code">default.nix</span> files; these set all dependencies and variables that are added to your terminal.</p>
```
yegortimoshenko: `nix-shell` is usually configured by `shell.nix` file, and `nix-build` is usually configured in `default.nix`, is that because of `nix-shell https://holochain.love` workaround? cc @thedavidmeister

start.html
```diff
@@ -312,10 +263,10 @@ <h5>This guide explains using NixOS with Vagrant/VirtualBox.</h5>
 
                                         <p>With nix-shell, you don’t need to worry about updating or uninstalling; when you enter the nix-shell, everything is the latest release and is then cleaned up when you exit.</p>
 
-                                    </div>
-                                </div>
-
-                                <br><h2 class="section-title">Build your first DNA</h2>
+                                        <br><h5>Editor</h5>
+                                        <p>In most cases you can just run your editor as normal. However if you are using an integrated developer envoronment or IDE that needs to communicate with the Holochain dependancies then you should launch it from inside the nix-shell.</p>
```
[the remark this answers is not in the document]
freesig: Thanks for the detail feedback @yegortimoshenko It's really helpful. I'll make these changes now.

start.html
```diff
@@ -312,10 +263,10 @@ <h5>This guide explains using NixOS with Vagrant/VirtualBox.</h5>
 
                                         <p>With nix-shell, you don’t need to worry about updating or uninstalling; when you enter the nix-shell, everything is the latest release and is then cleaned up when you exit.</p>
 
-                                    </div>
-                                </div>
-
-                                <br><h2 class="section-title">Build your first DNA</h2>
+                                        <br><h5>Editor</h5>
+                                        <p>In most cases you can just run your editor as normal. However if you are using an integrated developer environment or IDE that needs to communicate with the Holochain dependencies then you should launch it from inside the nix-shell.</p> 
+                                        <p><em>To do this just open your edit while you are in the nix-shell like: <span class="button-hide"><code>[nix-shell:</code></span><code>vim my_file.rs</code></em></p>
```
dhtnetwork: CHANGE from: edit ---> to editor

dhtnetwork:
"perhaps it should be more detailed and list how to launch all the common editors?"  

What will that look like? 

Do we support all editors? And what would providing a detailed list look like? Will the instructions be different. It looks it sound complicated. I say we go with the simple and see what feedback we are getting. We can consider it, but it's best to see how that would look like first before we go that route.

start.html
dhtnetwork: > CHANGE from: edit ---> to editor

dhtnetwork: when possible, remove the word "just".  It does not seem necessary.

freesig:
> This sounds complicated. 

I originally thought that it was just a matter of calling eg. `code lib.rs` from the nix-shell but vscode and other editors __don't__ automatically put themselves in the PATH so this is probably beyond a.

I'm happy to merge this now if everyone is happy

nix.html
thedavidmeister: @yegortimoshenko yes! it used to be two files but i merged them in a bit of a hack to get URL shell working

PR #76, opened 2019-10-11
Fix address

jamesray1:


pdaoust:
Thanks for that catch @jamesray1 !

PR #84, opened 2019-10-16
Add PyMdown Extras

pdaoust:
To support definition lists, fancy punctuation, etc. The annoying thing about this extension is that it comes with a bunch of things, and you can't really tease them apart. May require fine-grained review of the built site.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -187,3 +187,4 @@ markdown_extensions:
   - pymdownx.emoji
   - pymdownx.details
   - pymdownx.superfences
+  - pymdownx.extra


freesig:
Does it look ok when you run it locally? It should be pretty easy to remove this line. Once it's in develop we can PR to master and see what it looks like before publishing.

pdaoust:
I.... think it looks okay? I've been doing a side-by-side comparison of local vs live, and haven't seen anytihng yet. My biggest concerns were nested stuff -- e.g., code blocks within admonitions. So far those look great.

PR #127, opened 2019-11-16
Fix: (docs) conductorConfig => config

syzer:


freesig:
This has been fixed through a previous PR but that you so much for the PR :)

Issue #149, opened 2019-12-05
Catalina


freesig:
I'm opening this issue to investigate getting nix-shell to work on catalina. So far I can't find a working solution. It's not working for @dhtnetwork

freesig:
Potential solutions https://discourse.nixos.org/t/current-status-of-nix-on-macos-catalina/4286/6

maxnachamkin:
@freesig 

Try this.

https://github.com/NixOS/nix/issues/2925#issuecomment-564493970

erlend-sh:
Should eventually be fixed by https://github.com/NixOS/nix/pull/3212

dhtnetwork:
Yay, this has been resolved! Thank you for your help @freesig!! To resolve this issue, please check out these instructions: https://github.com/NixOS/nix/issues/2925#issuecomment-539570232

PR #155, opened 2019-12-16
Explain Catalina blocker

erlend-sh:
As pointed out in #149, Catalina might be partially blocked for a while.


dhtnetwork:
Hi @erlend-sh, have you been able to resolve this issue? I was able to resolve it today. So this may not be a global thing.  

A fix: https://github.com/NixOS/nix/issues/2925#issuecomment-539570232

If this is still a need let us know.

erlend-sh:
I haven’t tried the workaround myself yet but it should suffice. 👍 

Still need to point to that clearly in the docs though.

PR #181, opened 2020-01-17
hello me: fix for span variable renaming

e-nastasia:
Variable `span` is being renamed, but only in 1 out of 2 places it's being used. This commit fixes it.


pdaoust:
Closing; retired the old docs

PR #223, opened 2020-05-28
Hotfix: installation instructions for specific holonix version

pdaoust:


pdaoust:
okay, turns out there's a lot of hotfixes in here -- typos, content imiprovement, simplification of JS.

PR #234, opened 2020-06-23
Release -- simplified installation instructions

pdaoust:
* Ditch Windows/Vagrant
* Windows 10 users need to use WSL2
* Nix has an installer that works with Catalina
* The one-liner Nix install command has changed


pdaoust:
Fails CI because of a regression in HC .49 that causes the Simple Micro Blog test to fail. (links on agent ID don't propagate) Deploying anyway, with a warning at the top of that tutorial.

PR #256, opened 2021-01-29
New stuff: static snapshot of rustdoc

pdaoust:
Hand-picked, hand-compiled Rustdoc sites for these crates:

* `hdk3`
* `holochain_conductor_api` (for the admin and app APIs as well as the `ConductorConfig` struct)

Also included an orienting page for all of the above.


pdaoust:
And added a LOT of fixes to stuff that didn't make sense -- broken/obsolete URLs, etc.

PR #273, opened 2021-04-01
tone down RSM banner colour and wording

pdaoust:


pdaoust:
going to remove it instead

PR #276, opened 2021-04-01
Remove RSM guidance page

pdaoust:
It's not needed anymore, and in fact you can't even access it because the only link to it was on the RSM banner (removed in #274 ). I moved the only orphaned link -- the DNA build tutorial -- to the 'Resources' section. However, the content at that URL https://github.com/holochain/holochain/blob/develop/docs/build_tutorial.md is out of date and needs some love.


pdaoust:
Removed link to build tutorial; gonna have to put it in later.

PR #279, opened 2021-04-28
Big changes - move /docs/ to /

pdaoust:
- [x] removed old home page
- [x] removed markety pages
- [x] moved `/docs` to `/`
- [x] set up redirects
- [x] checked redirects
    - [x] from `/docs` to `/`
    - [x] from Redux URLs (guidebook, core concepts)
- [x] moved favicon, redirect page, and static Rustdoc to `/src` folder so it gets deployed


pdaoust:
Not ready to merge; guidebook redirect magically works on the Redux site but the core concepts don't

PR #289, opened 2021-07-14
Documentation drive base

Connoropolous:
Navigate the repository via the `documentation-drive` branch, if you want to add edits: https://github.com/holochain/docs-pages/tree/documentation-drive

This is the "main" branch for the sake of the documentation drive. We can see the documentation for the current state of this branch live at: https://deploy-preview-289--fervent-tesla-2a3da2.netlify.app/

- [ ] update images
- [ ] add individual sections of app architecture with linkable headings and list in the table of contents


Connoropolous:
For the 'application architecture' section... I came to realize this was *actually* the place to describe what I was starting to describe in 'intro to happ development'. Now that article can become like the same thing, but technical instead of explanatory. From the client through to zome. 

Please see the before and after of the 'application architecture', but here is the cliffs notes:
before: https://developer.holochain.org/concepts/2_application_architecture/
after: https://deploy-preview-289--fervent-tesla-2a3da2.netlify.app/concepts/2_application_architecture/
- make it rooted in a request from a client, and start with that request from the client
- start from the "outside in", instead of the "inside out", because the outside is where people have familiarity, so we won't lose them right away. 
- Don't use term 'hApp bundle' which may be reserved for Holo use case? There is 'hApp', but that refers to a bundle of DNAs only, no UI. Even the `launcher` is not conflating the two. 
- The graphics need to also integrate this 'no hApp bundle' distinction. Just make a 'hApp' a bundle of DNAs.
- Introduce importantly the layers of "Cells" (which fit into "slots")
- TODO: add an image of 'Cell' and 'hApp'

@pdaoust can you get in on this? @guillemcordoba thoughts?

guillemcordoba:
Uh haven't been able to look at this carefully, but I do know that the Holochain core team is in the middle of refactoring what a happ is, specially when paired with a UI... Wouldn't want to make any changes in the official docs until the dust settles there.

Connoropolous:
oh dear... eternally. 
"composability of hApps via ‘apps->dna->zomes’ is very confusing, and seems to always be in transition to something else thus not ever getting crystal clear" (https://forum.holochain.org/t/overall-thoughts-from-developers-using-holochain/6348)

Also, I really don't think that it should be assumed that user interfaces are web!

Connoropolous:
Can you let me know how who I should get in touch with about it? @guillemcordoba

guillemcordoba:
Yeah I understand, but there is a package containing a set of DNAs + web UIs that's really useful as consuming package for both the launcher and Holo (hApp devs only need to publish this package and it should work everywhere). That's my understanding, at least.

I think it's @zippy and @artbrock actually who are holding this transition.

Connoropolous:
Kk. Well, regardless, the bulk of the change is to reverse the order, and conceptualization, of the Application Architecture. To start from the "outside in" (an api call to your hApp) instead of the "inside out" (what is a zome?) so I'd like to hear your @guillemcordoba and @pdaoust feedback on that bit too. 

It's what inspired this, which is now a complement to this concept, as a tutorial/ example repo: https://github.com/holochain/happ-client-call-tutorial
The two would work very well hand in hand if these changes came in

pdaoust:
Merging and getting to ready to publish. Thanks to all for your work on this!!!!!

pdaoust:
(thought I'd leave the branch around in case we want to re-open for a new round of changes.)

PR #292, opened 2021-07-16
Added explanation for pinned default.nix files

guillemcordoba:
This has been one of the major points of confusion for new hApp devs I've seen.


src/install.md
```diff
@@ -170,6 +170,22 @@ It will always keep you up to date with the newest stable version of Holochain a
 
 Read through our [advanced installation guide](../install-advanced/) for tips and tricks on making your development environment easier to work with.
 
+## Setting up a repository with a pinned Holochain version
```
Connoropolous, line 173: "Setting up" might be less clear than it needs to be... this would be like "Using Holochain in a ..."

Connoropolous (changes requested):
Comment about the title of the section, but other than that this is AWESOME ... big step up on the whole 'version challenges' front

Connoropolous (approved):

PR #295, opened 2021-07-30
Post-docs-drive edits to Application Architecture

pdaoust:
Just massaging it a bit -- mainly to align the writing style with what's already there, but I also restructured a bit.

**Work in progress**; not done yet and in particular I haven't added any new images.


src/concepts/2_application_architecture.md
```diff
@@ -39,31 +39,63 @@ Holochain creates a ‘double membrane’ for each participant, bridging between
...
+A Holochain application or **hApp** allows a single user or **agent** to easily install and manage a suite of functionality, such as chat, accounting, or project management. A hApp is often made of multiple DNAs, each providing an aspect of functionality, and it has a **slot** for each.
+
+Some slots can be filled with multiple cells **cloned** from the same DNA. This is useful for creating separate spaces inside a single hApp --- individual project workspaces or private chat channels, for example.
+
+### Cell
+
+**Cells** occupy slots in a hApp. Each cell is a combination of a user's unique cryptographic key (their **agent ID**) and a running DNA. Within one user's instance of a hApp, all cells share the same agent ID.  Each cell acts as the user’s personal agent --- every piece of data that it creates or message it sends, it does so from the perspective of that agent.
```
Connoropolous, line 80: " every piece of data that it creates or message it sends, it does so from the perspective of that agent." this could be a link to 'source chain' in the glossary or something. I would say introduce the term, but omg there's so many terms to throw around :p

src/concepts/2_application_architecture.md
```diff
@@ -39,31 +39,63 @@ Holochain creates a ‘double membrane’ for each participant, bridging between
...
+
+Each cell stores its own agent history and is a member of a separate, isolated network within the hApp with its own data store.
+
+When a client calls a function in a hApp, it specifies the **cell ID**, which is the hash of the DNA plus the agent ID, along with the zome name, function name, and of course the input payload.
```
Connoropolous, line 84:
```suggestion
When a client makes a call to the API of a hApp, it has to specify the **cell ID**, which is the hash of the DNA plus the agent ID.
```

those other inputs can be left and introduced level by level?

src/concepts/2_application_architecture.md
```diff
@@ -39,31 +39,63 @@ Holochain creates a ‘double membrane’ for each participant, bridging between
...
 
-4. image here. Things known as **Cell**s occupy slots in a hApp. Since a hApp has to be uniformly connected to a specific agent (via its keys), that means that each slice of functionality of the hApp, like "project management", is running as the same agent (via its keys). Each Cell acts as a user’s personal agent — every piece of data that it creates or message it sends, it does so from the perspective of the agent. When we make a request from the client, we need to specify the id of a Cell. A Cell ID is a pairing of the *hash that identifies the bundle of code that implements the "project management"* (the DNA), and the *public key which represents the agent*. 
+The unique ID and properties can be changed either in a text editor, at installation time, or at cloning time. Due to the properties of cryptographic hashing, even the slightest alteration of either the code, properties, or unique ID will change the hash, forming a cell with an entirely new history and peer-to-peer network. This means that you should introduce changes very carefully.
```
Connoropolous, line 92:
```suggestion
The unique ID and properties can be changed either in a text editor, at installation time, or at cloning time. Due to the properties of cryptographic hashing, even the slightest alteration of either the code, properties, or unique ID will change the hash, forming a cell with an entirely new history and peer-to-peer network. This means that you should introduce changes very carefully. Software updates may use a pattern similar, but different, to traditional database migrations.
```

src/concepts/2_application_architecture.md
pdaoust: I know; soo soo many terms. Grappling with how to progressively introduce ideas without causing the reader to go madly off in all directions.

src/concepts/2_application_architecture.md
pdaoust: I also wonder if this is an implementation detail and should be pulled out entirely -- trying to be light on details and heavy on concepts. However... there is no guidebook yet, so the core concepts have to suffice. Probs what I'll do is move it to the very end, once all the bits have been introduced in detail.

src/concepts/2_application_architecture.md
pdaoust: Could you elaborate more on this? I feel like it's an important thing to bring up, and this is the right place to talk about it, but I'm having a hard time grasping what this pattern looks like.

Connoropolous: well, what we're learning with Acorn is that the handling of software updates, if they're to include DNA updates, are entirely up to us as the software developers to define, with little to no guidance from Holochain. We currently are providing end users with JSON data dump exports of their projects, that they can re-import into a new version... then the version will have to be responsible for realizing what the schema is they're coming from so that it can import it correctly.

src/concepts/2_application_architecture.md
Connoropolous: I was thinking that at the end of this you can just link off to the (now even more flushed out) happ-client-call-tutorial which would give you details.

src/concepts/2_application_architecture.md
Connoropolous: Not posing this is as a criticism ^

src/concepts/2_application_architecture.md
```diff
@@ -76,17 +96,18 @@ Functional components and architectural layers both enjoy clean separation. You
...
+* Zomes don’t maintain any in-memory state between calls; state is maintained in the history of the cell that contains the zome.
 * Persistence logic and core business logic are mixed together in your DNA, because at its heart, Holochain is a framework for data validation. However, you usually don’t need a lot of your business logic in your DNA—just enough to encode the ‘rules of the game’ for your application.
 * As with microservices, Holochain lends itself to combining small, reusable components into large applications.
 
 ## Learn more
 
-* [Holochain: Reinventing Applications](https://medium.com/holochain/holochain-reinventing-applications-d2ac1e4f25ef)
-* [The Holo vision: Serverless 2.0](https://medium.com/holochain/the-holo-vision-serverless-2-0-c0b294e753ba)he foundation of Holochain is simple, but the consequences of our design can lead to new challenges. H
+If you want to go from theory to practice, you can learn hands-on how applications talk to hApps by reading and following the [hApp client call tutorial](https://github.com/holochain/happ-client-call-tutorial), which has examples in Typescript and Rust.
```
Connoropolous, line 113: You can see in my git commit my rationale... 
I looked in to those links and they are very lofty and off topic. I think it'd be good to take this chance to ground the technical reader.

src/concepts/2_application_architecture.md
pdaoust: I don't want to get too far into the weeds re: migration -- I see this as a job for a future Guidebook -- but I do want to mention something as you say. I left it open-ended re: how to implement migration; the commit isn't showing up in this PR yet though 🤷

src/concepts/2_application_architecture.md
```diff
@@ -7,67 +7,68 @@ Applications built with Holochain are highly **modular**. This makes it easy to
...
+
+### Zome
+
+![](../../img/concepts/2.5-zome.png)
+
+The executable code modules in a DNA are called **zomes** (short for chromosomes), each with its own name like `profile` or `chat`. The zomes define the core business logic in a DNA, exposing their functions to the conductor. Some of these functions are 'hooks' called automatically by Holochain, such as an initialization function or validation functions related to data types defined in the zome.
 
-Functional components and architectural layers both enjoy clean separation. You can combine, augment, or replace existing pieces. This gives you a lot of flexibility and can even empower your users to take ownership of their experience.
+Other functions are invented by the developer, have arbitrary names, and define the zome’s public API. The conductor [makes this API available](../8_calls_capabilities/) to other zomes within the DNA, other DNAs within the hApp, and, as mentioned earlier, clients running on the participant's machine and other agents on the DNA's network. The developer can give a participant the ability to control access to their cell's API via [capabilities](../8_calls_capabilities/).
```
pdaoust, line 71: @Connoropolous do you think it's long-winded to introduce all these details here? I'm wondering if this should be saved for the Guidebook -- at the same time, devs need to know how their code can be called and probably want to know how access can be controleld too.

src/concepts/2_application_architecture.md
```diff
@@ -7,67 +7,68 @@ Applications built with Holochain are highly **modular**. This makes it easy to
...
 
-</div>
+The DNA can also contain metadata: a name, description, unique ID, and **properties**. The unique ID and properties can be changed either in a text editor or at installation time. The unique ID, on the other hand, can be changed to **clone** a DNA, creating a new cell with identical functionality but an entirely separate history, network, and shared database. The properties, on the other hand, can also be changed to clone a DNA, but also direct the executable code to change the new cell's runtime behavior (similar to configuration parameters). 
+
+In fact, even the slightest alteration of any part of the DNA will do this. Consider source code changes carefully, because each modification will create a new DNA with a new cells interacting in a separate network. This may require some sort of migration strategy to move or access data between old and new cells, which means that you should introduce changes carefully.
```
pdaoust, line 63: @Connoropolous This is my latest iteration of the commentary on migration. What do you think? Leaving implementation details unspecified/open-ended (I sorta think special migration zome functions + bridging are the answer, but I suspect changing a DNA means you need to migrate to a totally new hApp and I don't know if one hApp can call another's functions).

pdaoust:
I think I'm finished on the text; just need to change those images!

src/concepts/2_application_architecture.md
pdaoust: Sounds totally reasonable to me. Thanks for checking those articles.

src/concepts/2_application_architecture.md
Connoropolous: looks good

src/concepts/2_application_architecture.md
Connoropolous: Good, but there's repetition of "introduce changes carefully", maybe once more than necessary

Connoropolous:
The images are hilarious and great!

Connoropolous:
extremely useful that you can now link to each specific heading... hyperlink that all over the place

Connoropolous:
Also very happy this keeps to the pure Holochain ontology which limits to 'hApp' and not the extrapolated 'happ bundle' which includes a pontential web based UI. Place for that will be elsewhere. Much much clearer and a good context fit

src/concepts/2_application_architecture.md
pdaoust: oh! good catch

pdaoust:
@Connoropolous Glad you like the new images! Was worried I was getting carried away with the cell/DNA/zome imagery there.

Re: hApp and hApp bundle... at some point it might be good to make a mention of different 'officially sanctioned' ways of creating/deploying a client... some day.

pdaoust:
Closed this for now; please create an issue or PR against `develop` for any feedback. Thanks!

PR #298, opened 2021-08-01
go live with the minor edits from docs drive

Connoropolous:


Connoropolous:
@pdaoust can you approve of these so we can merge them? there's lots of folks coming through holochain dev docs right now, it seems like

Connoropolous:
Actually @pdaoust I'm going to take a little bit of a 'move fast' on this one, knowing your next work day is in a few days. Let's put it this way, if you come up with anything catastrophic from these changes then we can revert :p

PR #311, opened 2021-11-23
switch to stable holonix and add cachix instructions

steveeJ:
This goes alongside https://github.com/holochain/holonix/pull/258 and the previously merged https://github.com/holochain/holonix/pull/253.


pdaoust (changes requested):
Ugh, GitHub! :( All your edits look great; I just removed the word 'built and' and a non-existent URL from the advanced install guide. It's the bits that you didn't edit that I want to suggest changes on, and there's no facility for that! Maybe I can just list them here:

* Line 73 says "Note that nix will have to compile all the binaries from source, so it'll take a long time"; could you change it to "Note that, on first run, nix will have to download and cache all the packages for this version, and even compile them if the version is old enough. This might take a long time."

Okay, looks like there was only one. Thanks very much for taking these edits @steveeJ !

steveeJ:
thanks for the review @pdaoust! i integrated your suggestions. PTAL

pdaoust (approved):

PR #316, opened 2022-01-31
Simplified installation instructions

jost-s:
I've simplified the installation instructions and broken down the section on using Holochain dev tools into using ad-hoc shells and creating `default.nix` files. @pdaoust We're working towards a simple process of upgrading the pinned Holochain version.

@steveeJ Please review. I need your help with the `niv` command.
Looking through our conversations I saw another form of `default.nix`:
```nix
let
  holonixPath = builtins.fetchTarball "https://github.com/holochain/holonix/archive/$(git ls-remote https://github.com/holochain/holonix main | grep -oP '[0-9a-z]{40}').tar.gz";
  holonix = import (holonixPath) {
    holochainVersionId = "v0_0_120";
  };
  nixpkgs = holonix.pkgs;
in nixpkgs.mkShell {
  inputsFrom = [ holonix.shell ];
  packages = [
    # additional packages go here
  ];
}
```
Would that be even easier than running `niv`, because one only needs to increase the version number?

### DIFF
--- a/src/install.md
+++ b/src/install.md
@@ -199,6 +199,7 @@ in nixpkgs.mkShell {
   ];
 }
 ```
+
 #### Upgrading Holochain in default.nix
 
 All you need to do to upgrade the version of Holochain in your project is run
@@ -206,6 +207,7 @@ All you need to do to upgrade the version of Holochain in your project is run
 ```bash
 nix-shell --run "niv update"
 ```
+
 ### Advanced usage
 
 Read through our [advanced installation guide](../install-advanced/) for tips and tricks on making your development environment easier to work with, or what to do in case you need to work offline.
--- a/src/install.md
+++ b/src/install.md
@@ -194,15 +194,20 @@ in nixpkgs.mkShell {
   inputsFrom = [ holonix.main ];
   packages = with nixpkgs; [
     niv
-
     # any additional packages needed for this project, e. g. Nodejs
   ];
 }
 ```
 
-#### Upgrading Holochain in default.nix
+#### Upgrading Holochain referenced in default.nix
 
-All you need to do to upgrade the version of Holochain in your project is run
+If you want to be in a position to automatically upgrade your project to the most recent version of Holochain, you need to initalize a tool called `niv` after setting up your project.
+
+```bash
+nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-21.11.tar.gz -p niv --run "niv init && niv drop nixpkgs && niv drop niv && niv add -b main holochain/holonix"
+```
+
+`niv` serves as an easy way to manage dependency. Once it's initalized, all you need to do to upgrade Holochain to the most recent version in your project is run
 
 ```bash
 nix-shell --run "niv update"
--- a/src/install.md
+++ b/src/install.md
@@ -188,7 +188,7 @@ If you want to automatically upgrade your project to the most recent version of 
 nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-21.11.tar.gz -p niv --run "niv init && niv drop nixpkgs && niv drop niv && niv add -b main holochain/holonix"
 ```
 
-Executing this command creates a new folder named `nix` with 2 files that `niv` needs to retrieve the revision of the latest Holochain version.
+Executing this command creates a new folder named `nix` with 2 files that `niv` needs to retrieve the revision of the latest Holonix version.
 
 Next you need to add a `default.nix` file as a Nix configuration of your development environment, including Holochain as a dependency.
 This is a minimal `default.nix` file with a specific version of Holochain:
@@ -211,9 +211,25 @@ in nixpkgs.mkShell {
 
 Now you can enter the Nix shell with your development environment by running the command `nix-shell`. Once it's finished you have the Holochain commands at your disposal.
 
-#### Upgrading to the latest version of Holochain
+> niv initializes and updates to the latest revision of the Holonix repository. As every revision contains configurations for several previous versions of Holochain, you need to explicitly define the exact version of Holochain you want to use. In other words, niv does not set a Holochain version; it's defined in `default.nix`.
 
-When the time has come to upgrade your hApp to the latest version of Holochain, you first have to set the version id in `default.nix`. E. g.
+#### Upgrading the Holochain version
+
+When the time has come to upgrade your hApp to a newer version of Holochain, there are 3 steps to follow:
+
+1. Update the Holonix revision using `niv`:
+
+```bash
+nix-shell --run "niv update"
+```
+
+2. Run `hn-versions` to see which versions of Holochain are available:
+
+```bash
+nix-shell --run "hn-versions"
+```
+
+3. Set the `holochainVersionid` accordingly:
 
 ```nix
 ...
@@ -223,13 +239,9 @@ When the time has come to upgrade your hApp to the latest version of Holochain, 
 ...
 ```
 
-Then all you have to do is execute
+Now you can `exit` the Nix shell and re-enter it via `nix-shell`. The updated version of Holochain will be downloaded and made available in the resulting Nix shell.
 
-```bash
-nix-shell --run "niv update"
-```
-
-After the update completes, you can exit the Nix shell by running `exit` and re-enter it via `nix-shell`. The updated version of Holochain will be downloaded and made available in the result Nix shell.
+> Keep in mind that the Holonix repo includes Nix configurations for the last ~ 5 versions of Holochain. That means that if you keep updating its revision using `niv`, you will have to augment the Holochain version id in `default.nix` sooner or later too.
 
 ### Advanced usage
 
--- a/src/install.md
+++ b/src/install.md
@@ -171,17 +171,26 @@ It will always keep you up to date with the newest stable version of Holochain a
 
 Holochain is currently in rapid development, which means newer versions introduce new features and breaking changes. This means that it's likely that the version that you get with `nix-shell https://holochain.love` won't always work with existing hApp repositories or even breaks a hApp you were working on.
 
-To solve this, repositories of hApp projects can use Nix to pin the appropriate Holochain version to work well for that hApp. To do this, the project needs to have a `default.nix` file in the root folder of the repository. You must not be inside the `nix-shell` provided by https://holochain.love. Instead, simply navigate to the folder with the `default.nix` file and run:
+To solve this, hApp projects can use Nix to pin a compatible Holochain version. The project needs to have a `default.nix` file in the root folder of the repository. Don't run this from inside the `nix-shell` provided by https://holochain.love. Instead, simply navigate to the project's root folder where the `default.nix` file needs to be and run:
 
 ```bash
 nix-shell
 ```
 
-This command looks for a `default.nix` file in the current folder and will create the environment as specified.
+This command looks for a `default.nix` file in the current folder and will create the specified environment.
 
-#### Example default.nix file
+#### Initialize a project for easy Holochain version upgrades
 
-This is a minimal `default.nix` file to use a specific version of Holochain:
+If you want to automatically upgrade your project to the most recent version of Holochain, you need to initalize a tool called `niv` for your project.
+
+```bash
+nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-21.11.tar.gz -p niv --run "niv init && niv drop nixpkgs && niv drop niv && niv add -b main holochain/holonix"
+```
+
+Executing this command creates a new folder named `nix` with 2 files that `niv` needs to retrieve the revision of the latest Holochain version.
+
+Next you need to add a `default.nix` file as a Nix configuration of your development environment, including Holochain as a dependency.
+This is a minimal `default.nix` file with a specific version of Holochain:
 
 ```nix
 let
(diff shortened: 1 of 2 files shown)


src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
-It will always keep you up to date with the newest stable version of Holochain and the dev tools. If you need to work offline, read the [advanced installation guide](#keeping-everything-local-working-offline).
-
-### Going further
-
-Read through our [advanced installation guide](../install-advanced/) for tips and tricks on making your development environment easier to work with.
-
-## Using Holochain in a repository with a pinned Holochain version
-
-Holochain is currently in rapid development, which means newer versions introduce new features and breaking changes. This means that it's likely that the version that you get with `nix-shell https://holochain.love` won't always work with existing hApp repositories.
+It will always keep you up to date with the newest stable version of Holochain and the dev tools.
```
pdaoust, line 168:
```suggestion
It will always keep you up to date with the newest stable version of Holochain and the dev tools.

```

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
-If you are trying to set up a repository that comes with a `default.nix` file, you must not be inside the `nix-shell` provided by https://holochain.love. Instead, simply navigate to the folder with the `default.nix` file and run:
+To solve this, repositories of hApp projects can use Nix to pin the appropriate Holochain version to work well for that hApp. To do this, the project needs to have a `default.nix` file in the root folder of the repository. You must not be inside the `nix-shell` provided by https://holochain.love. Instead, simply navigate to the folder with the `default.nix` file and run:
 
 ```bash
-nix-shell .
+nix-shell
 ```
+This command looks for a `default.nix` file in the current folder and will create the environment as specified.
+
+#### Example default.nix file
```
pdaoust, line 180:
```suggestion
#### Example default.nix file

```

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
+This is a minimal `default.nix` file to use a specific version of Holochain:
```
pdaoust, line 181:
```suggestion
This is a minimal `default.nix` file to use a specific version of Holochain:

```

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
+    holochainVersionId = "v0_0_124"; # specifies the Holochain version
+  };
+  nixpkgs = holonix.pkgs;
+in nixpkgs.mkShell {
+  inputsFrom = [ holonix.main ];
+  packages = with nixpkgs; [
+    # any additional packages needed for this project, e. g. Nodejs
+  ];
+}
+```
```
pdaoust, line 195:
```suggestion
```

```

File: src/install.md [hunk 5]
```diff
@@ -169,41 +154,53 @@ holochain --version
...
+#### Upgrading Holochain in default.nix
```
pdaoust, line 196:
```suggestion
#### Upgrading Holochain in default.nix

```

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
+All you need to do to upgrade the version of Holochain in your project is run
```
pdaoust, line 197:
```suggestion
All you need to do to upgrade the version of Holochain in your project is run

```

pdaoust: (this is a very cool new tool, by the way. Is it useful to explain what it does -- I assume it updates your `default.nix` and `Cargo.toml` files? `Cargo.lock` too? Would it be necessary to also mention that breaking changes are your own responsibility to fix, or is that something devs will assume?)

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
+```bash
+nix-shell -p niv --run "niv update ..."
+```
```
pdaoust, line 200:
```suggestion
```

```

File: src/install.md [hunk 8]
```diff
@@ -119,33 +107,30 @@ nix-shell --version
...
-nix run -f https://cachix.org/api/v1/install cachix -c cachix use holochain-ci
+nix-env -iA cachix -f https://cachix.org/api/v1/install
+cachix use holochain-ci
 ```
 
-If does not work your version of Nix might be outdated. In this case please try steps 2 and 3 from [the documentation on cachix.org](https://app.cachix.org/cache/holochain-ci).
-
-### Load Holonix
+## Using the Holochain dev tools
+You can use the Holochain dev tools in two ways, either by creating a nix file which specifies the Holochain version or by using a link that points to a current version. Usually Holochain projects define a specific Holochain version that they're running on and compatible with, just like Rust crates or NPM packages are included in a project with a particular version. If you  want play around with Holochain and the dev tools, you can use the link with an ad-hoc nix-shell.
```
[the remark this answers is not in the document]
pdaoust: Oh! On re-reading I get what the 'link' method is -- that's the same as holochain.love, right? I was picturing a symlink. In that case, how about this:

```suggestion
You can use the Holochain dev tools in two ways, either by [creating a nix file](#using-holochain-with-a-pinned-holochain-version) which specifies the Holochain version or by using a URL that points to the most recent release. Usually Holochain projects define a specific Holochain version that they're running on and compatible with, just like Rust crates or NPM packages are included in a project with a particular version. If you  want play around with Holochain and the dev tools, you can use the URL with an ad-hoc nix-shell.
```

pdaoust (commented):
Thanks very much for updating this -- it's not just exciting to learn about the new upgrade tool and have an explicit example of pinning, but also to see the improvements in Nix which make a lot of our installation details unnecessary now.

steveeJ (commented):
thank you for getting this rolling!

please consider the comments i made :slightly_smiling_face:

src/install.md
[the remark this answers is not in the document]
jost-s: Yeah, definitely, thanks for the improvement, @pdaoust.
And sure, let's add the whitespace. I didn't take the time to figure out setting up a local dev server to view the generated site, so all your typesetter view's touches are much appreciated 😀

src/install.md
```diff
@@ -169,41 +154,53 @@ holochain --version
...
-nix-shell .
+nix-shell
 ```
+This command looks for a `default.nix` file in the current folder and will create the environment as specified.
+
+#### Example default.nix file
+This is a minimal `default.nix` file to use a specific version of Holochain:
+```nix
+let
+  holonixPath = (import ./nix/sources.nix).holonix; # points to the current state of the Holochain repository
```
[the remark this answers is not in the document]
jost-s: What a complicated one-liner. Okay, so that command generates a `default.nix` of this kind?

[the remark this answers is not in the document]
steveeJ: no it doesn't generate a `default.nix` file at all. it creates the files in `nix/` that the `default.nix` you give in the example relies on.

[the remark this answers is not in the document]
jost-s: So you need to run this command and create/copy a default.nix file and then updating the version of Holochain to the current version of branch `main` works by issuing `nix-shell --run "niv update"`?

src/install.md
pdaoust: weird, this was not what I suggested. GitHub is doing weird things... what I wanted was a newline before and after the code block.

src/install.md
jost-s: I've added the whitespace, @pdaoust.

src/install.md
jost-s: Whitespace added.

src/install.md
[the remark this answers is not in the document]
steveeJ: @jost-s yes that's correct. the nix code in your example references the file `./nix/sources.nix` which is generated by `niv`.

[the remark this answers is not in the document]
jost-s: @steveeJ Thanks. I've updated that section and added the `niv` init. Please check that section again.

Two more questions:
1. I've tried to upgrade the Holochain dependency in `default.nix` with these two steps. I. e. I created a folder, added the `default.nix` file, then ran the `niv` init command and then executed the `niv` update command. The commands were successful, but the `default.nix` file wasn't changed. So I must be missing something.
2. Please check my initial comment:

> @steveeJ Please review. I need your help with the `niv` command. Looking through our conversations I saw another form of `default.nix`:
> 
> ```nix
> let
>   holonixPath = builtins.fetchTarball "https://github.com/holochain/holonix/archive/$(git ls-remote https://github.com/holochain/holonix main | grep -oP '[0-9a-z]{40}').tar.gz";
>   holonix = import (holonixPath) {
>     holochainVersionId = "v0_0_120";
>   };
>   nixpkgs = holonix.pkgs;
> in nixpkgs.mkShell {
>   inputsFrom = [ holonix.shell ];
>   packages = [
>     # additional packages go here
>   ];
> }
> ```
> 
> Would that be even easier than running `niv`, because one only needs to increase the version number?

Is that so?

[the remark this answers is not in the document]
steveeJ: > 1. (...). but the default.nix file wasn't changed. So I must be missing something.

yes, none of these commands touch the `default.nix`. i tried to explain this in a previous comment here but it disappeared. `niv update` generates the `nix/sources.nix` based on the `nix/sources.json` file, and it does nothing else. by referencing the generated files in the `default.nix` we have access to the latest holonix revision. if the `default.nix` file specifies the `holochainVersionId = "v...."` argument to the holonix import, it will pin holochain to that version regardless of the holonix update. if that argument is either omitted or names a moving target such as `main` or `develop`, it will move with holonix. if omitted, the default on the `main` branch of holonix is `main`, and the default on the develop branch of holonix is `develop`. i hope this isn't surprising :laughing:

IMO the only thing we're missing to make this process smooth is a command that prints a list of available holochain packages and which ones are cached. if you like you can watch me implement it while i explain to you what i'm doing. it might help you understand the connections.

> 2. Would that be even easier than running niv, because one only needs to increase the version number?

this example works once, when generating the file. you probably took this from between an `cat <<<EOF (...) EOF`, which will substitute the shell command for its output. this will not help with subsequent updates.

jost-s:
Speaking with @steveeJ about the installation and upgrade of Holochain as part of `default.nix`, we realized that it's still pretty complex for newcomers and can easily be improved by providing a command that generates this `default.nix` and initializes `niv` so that upgrading is simpler.

Additionally I think that the one-liner `nix-shell https://holochain.love` is of little use for real world projects. Having it in the instructions first suggests that it's the easier way, but as we say there it's not reliable and may break unforeseeably. Two different options raise the complexity of the installation process, so ideally we have only one.

This lead us to a discussion about a minimal Holochain project scaffolding. We'll explore that possibility first and put this PR on hold.

steveeJ:
@jost-s 

the remaining UX item i describe above:

> IMO the only thing we're missing to make this process smooth is a command that prints a list of available holochain packages and which ones are cached

is addressed by https://github.com/holochain/holonix/pull/285

pdaoust (approved):
LGTM!

pdaoust:
oh, just reading upwards and saw that you're going to put the PR on hold. However, I do think it adds value right away and isn't any less helpful than the existing documentation. Even if it's not the ideal process, I still think it should get published for now.

jost-s:
@pdaoust I agree. We might complete this "advanced simplification" next week, so do you think we should publish now and then again in 1-2 weeks time?

pdaoust:
Sorry @jost-s I never saw your reply until now. Yes, I think we should deploy this as part 1 of 2. @steveeJ do you agree too?

jost-s:
@pdaoust He does here https://forum.holochain.org/t/dev-tools-install-fails-during-cachix-install/7910/2

jost-s:
@steveeJ I'm able to initialize this update mechanism with `niv`.

The command `nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-21.11.tar.gz -p niv --run "niv init && niv drop nixpkgs && niv drop niv && niv add -b main holochain/holonix"` however produces errors that "xx doesn't exist" for the `niv drop xx` commands. When I remove them, the command works.

After successfully running it, I can enter the nix shell that is pointing to `nix/sources.nix` and also execute `niv update`.

Is that fine or do we need the `drop` part?

steveeJ:
> The command `nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixos-21.11.tar.gz -p niv --run "niv init && niv drop nixpkgs && niv drop niv && niv add -b main holochain/holonix"` however produces errors that "xx doesn't exist" for the `niv drop xx` commands. When I remove them, the command works.

looking through your command the `niv drop nixpkgs` isn't needed as it's not added. but i think we need `niv drop niv` because `niv init` by default pulls in `niv` as a source, and we don't need that.

jost-s:
It works now as is. What I did before was that I tried to add `--pure` but that failed and then then non-pure command didn't work anymore, either.

Starting from scratch, the command above works fine.

I can also change the version number in `default.nix` and then run `niv update` successfully. When I then enter the nix-shell again and do `hn-introspect`, I get the updated holochain version.

All good it seems. I'm making the changes to the text and then we can merge this.

jost-s:
@steveeJ I've corrected the part that talks about using niv. Please review it.

1. I have only written that you can use niv to upgrade to the latest version, not to any other particular version. In my experiments that's what it did, regardless of which version I set in `default.nix`. Is that correct?
2. I wonder why I didn't have the command `hn-versions`? Should it not be included by now?

src/install.md
```diff
@@ -169,40 +155,85 @@ holochain --version
...
+  ];
+}
+```
+
+Now you can enter the Nix shell with your development environment by running the command `nix-shell`. Once it's finished you have the Holochain commands at your disposal.
+
+#### Upgrading to the latest version of Holochain
 
-If you are trying to set up a repository that comes with a `default.nix` file, you must not be inside the `nix-shell` provided by https://holochain.love. Instead, simply navigate to the folder with the `default.nix` file and run:
+When the time has come to upgrade your hApp to the latest version of Holochain, you first have to set the version id in `default.nix`. E. g.
```
steveeJ, line 216: i would say that the
1. first thing to do is `niv update`
2. then run `hn-versions` to see what versions are available
3. and then set the `holochainVersionid` accordingly

steveeJ (commented):
> I wonder why I didn't have the command hn-versions? Should it not be included by now?

it was supposed to be there. after double-checking i noticed that i didn't add it to the shell automatically, only as a package in the repo. [this pr changes that](https://github.com/holochain/holonix/pull/319/files).

src/install.md
jost-s: Yes, now that `hn-versions` is available, I've changed it so.

jost-s:
@steveeJ One last review, please.

steveeJ (commented):
yay, almost done :raised_hands:

PR #322, opened 2022-06-17
Refactor/rename element and header

jost-s:
### Changed

- replaced all occurrences
  - "element" -> "record"
  - "header" -> "action"

PR #324, opened 2022-06-22
rename element→record, header→action in images

pdaoust:
@jost-s here's the images with renamed text; I'm requesting to merge them into your PR #322 branch so they can all land in `master` at the same time.

(side note: we really should rename `master` to `main`; I hear it's not as simple as just `git branch -m` though)


jost-s (approved):
all good

PR #332, opened 2022-09-11
Remove happ-build-tutorial and happ-client-call-tutorial references, since they're "deprecated"

Connoropolous:
@zippy deprecated happ-build-tutorial so I did the same for happ-client-call-tutorial and so removing them from being linked from developer.holochain.org

Also suggested to @zippy that happ-build-tutorial be unpinned from the Holochain Github org


jost-s (approved):
LGTM

PR #337, opened 2022-11-21
add 'remote signal' to glossary, modify definition of 'signal'

pdaoust:


src/glossary.md
```diff
@@ -576,6 +576,10 @@ A [remote procedure call](#remote-procedure-call) that one agent's [cell](#cell)
 1. A call that a [client](#client) makes to a [zome function](#zome-function) or [conductor admin API](#conductor-admin-api) function over a local socket interface.
 2. A [remote call](#remote-call) between [agents](#agent) in a [network](#network).
 
+#### Remote signal
+
+A [signal](#signal) that one [cell](#cell) sends to another cell within a [network](#network) that they belong to. The receiver's cell processes the signal by implementing a special [zome function](#zome-function) with the name `recv_remote_signal`.
```
matthme, line 581: Maybe one could add another line explaining that it is not persisted and "fire and forget", something like this:
"Remote signals are not persisted to the [DHT](#DHT) and are "fire and forget", meaning that if they don't reach their recipient, they are lost and there will be no attempt to retry."

pdaoust: sure, thanks for the suggestion 👍🏼

pdaoust: I added it to 'signal' instead, cuz it's true of local signals too.

jost-s (approved):

src/glossary.md
```diff
@@ -610,7 +614,7 @@ A process of reducing the processing and storage load of individual [nodes](#nod
 
 #### Signal
 
-A message emitted by a [cell](#cell), meant for a [client](#client) to receive and act upon.
+A message emitted by a [cell](#cell) for another piece of code to act on, either a locally running [client](#client) or a [remote](#remote-signal) cell within the same [network](#network). Signals are 'fire-and-forget'; they are not persisted to any [source chain](#source-chain) or the [DHT](#dht) and will be lost and not retried if they don't reach the intended recipient(s).
```
jost-s, line 617:
```suggestion
A message emitted by a [cell](#cell) for a [client](#client) within the same [network](#network) to act on. Signals can be emitted and received within the same cell, or through a [remote signal](#remote-signal) be sent from one cell to other agents. All signals are 'fire-and-forget'; they are not persisted to any [source chain](#source-chain) or the [DHT](#dht) and will be lost and not retried if they don't reach the intended recipient(s).
```

jost-s (approved):
Here's a suggestion for the signal text. Signals need to be handled by clients in any case, whether it's a same-cell signal or a signal to other agents.

matthme (approved):

src/glossary.md
pdaoust: Evidently I'm not receiving GitHub updates, sorry I let this thread go dead. While another cell on the network is technically a client, I've been reserving that word for things that live outside of Holochain. I'm happy to have my mind changed, but it'll have to be a larger project -- lots of edits to the Core Concepts, etc.

pdaoust:
Closing this PR, as I've merged the changes into a separate brancht that includes a lot of other changes to RSM since I last touched this website.

PR #338, opened 2022-12-13
Update happ-setup.md

guillemcordoba:


matthme:
Just had a quick look at it. When comming to this page directly from the "Get started" page via the "Project Structure" button, it might be confusing why there should be a "forum" folder in every holochain project. Maybe it would make sense to make the connection that this is the structure that results from "hc scaffold example forum".
image

PR #341, opened 2022-12-14
docs(references): add lair to binaries

jost-s:


Connoropolous:
I would tweak the wording... it reads a bit like a dictionary definition that uses the word to define itself. Ideally one could glean what the purpose of the thing is from reading this snippet.

PR #348, opened 2023-01-25
remove outdated tips

jost-s:
I can't tell what will remain valid after the Nix dev shell refactoring is complete. For the time being I suggest what's in here.


marycamacho (commented):
I am approving this simplification. If we want to add more detail in the next developer portal version we can consider it in light of the new information architecture and the feedback and questions coming in from devs after the beta release.

marycamacho (approved):

PR #361, opened 2023-02-09
quick-start: add instructions to increase chances of successful cache setup; use `nix run`

steveeJ:
we don't have to merge all of it or as it is, but this captures some of the feedback and adds consistency


steveeJ:
the equivalent of these suggestions have been merged already :+1:

PR #363, opened 2023-02-09
Set up redirect for the quick-start

pdaoust:
There are a lot of inbound links to `/install/` in the wild, so it's important to 301 them to the right spot.


pdaoust:
I also added one for core concepts, since the main page has been removed and there's an awful lot of inbounds to `/concepts/` too.

PR #366, opened 2023-02-10
quick-start: replace nix-shell with nix

steveeJ:
this is not meant to be merged just yet. i'll be updating it and ask for testing eventually

- [ ] clarify whether we want to rely on the scaffolding to make ~/.config/nix/nix.conf entry or instruct users to do so @pdaoust 
- [ ] clarify whether we want to show usage of `hc scaffold` rather than directly using the `hc-scaffold` package, as this could be a source of confusion
- [ ] ask for testers

### DIFF
--- a/src/pages/quick-start/index.md
+++ b/src/pages/quick-start/index.md
@@ -33,19 +33,20 @@ After installing Nix, close the terminal and open a new one.
 Check that Nix is correctly installed:
 
 ```bash
-nix-shell --version
+nix --version
 ```
 
 You should see something like:
 
 ```bash
-nix-shell (Nix) 2.13.2
+nix (Nix) 2.13.2
 ```
 
-Run the following command to set up the cache for precompiled Holochain executables:
+Run the following commands to set up the cache for precompiled Holochain executables:
 
 ```bash
-sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix --extra-experimental-features nix-command --extra-experimental-features flakes -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
+export NIX_CONFIG="extra-experimental-features = nix-command flakes"
+sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix  -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
 ```
 
 ## Scaffold Your First Holochain App
@@ -53,7 +54,7 @@ sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix --extra-exp
 Type the following at the command line:
 
 ```bash
-nix-shell https://holochain.love --run "hc scaffold example forum"
+nix run github:holochain/holochain#hc-scaffold -- example forum
 ```
 
 When prompted, select the UI framework you prefer.
--- a/src/pages/get-building/index.md
+++ b/src/pages/get-building/index.md
@@ -14,7 +14,7 @@ For this example we will scaffold the canonical example, a To-do hApp.
 
 Run the holochain scaffolding tool by typing in your terminal:
 ```bash
-nix-shell https://holochain.love --run "hc scaffold web-app"
+nix run github:/holochain/holochain#hc-scaffold -- web-app"
 ```
 You should then see: 
 ```text


jost-s:
> clarify whether we want to rely on the scaffolding to make ~/.config/nix/nix.conf entry or instruct users to do so @pdaoust

I think definitely not. I'd make this the first step after nix installation, to add these commands to the user config.

> clarify whether we want to show usage of hc scaffold rather than directly using the hc-scaffold package, as this could be a source of confusion

I prefer not using the package, because it requires more knowledge about the workings of Nix. It's easier to comprehend that the nix commands create some dev environment in which Holochain tools are available and can then be called, instead of understanding the composition of Holonix and knowing that there's a package called hc-scaffold and that the main binary of that package is called.

It's consistent with what developers do once they have their scaffolded flake. They enter a dev shell and have several commands available to them, not running `nix run` command whilst knowing which package to call for which command.

steveeJ:
> I prefer not using the package, because it requires more knowledge about the workings of Nix. It's easier to comprehend that the nix commands create some dev environment in which Holochain tools are available and can then be called, instead of understanding the composition of Holonix and knowing that there's a package called hc-scaffold and that the main binary of that package is called.
> 
> It's consistent with what developers do once they have their scaffolded flake. They enter a dev shell and have several commands available to them, not running `nix run` command whilst knowing which package to call for which command.

this makes we want to put all the quick-start instructions behind a script so we abstract them away entirely. initializing a shell with _all_ of holonix just to run the scaffolding is inefficient because it installs all other components too. then we'd want to be more careful that the scaffolded and the initial shell are using the same revisions, which seems cumbersome.

jost-s:
You could add a dev shell only for scaffolding, but I think it's not all that important after all and you can keep the `nix run` command.

A script is supposed to be introduced anyway, isn't it?

steveeJ:
> A script is supposed to be introduced anyway, isn't it?

i haven't heard any technical objections to it. i think the next step is to write it and prepare a PR and ask a few people to test it.

steveeJ:
> You could add a dev shell only for scaffolding

it would also have to include `hc` so that the command would actually be `hc scaffold`, hence my hesitation as `hc` is part of the `holochain` package which ships a few more binaries.

pdaoust:
> It's consistent with what developers do once they have their scaffolded flake. They enter a dev shell and have several commands available to them, not running nix run command whilst knowing which package to call for which command.

This raises an interesting question -- should `hc-scaffold` even be available from a scaffolded flake? Seems to me that that would just create some sort of confusing recursive thing, where devs don't even realise that they should be asking themselves "if I scaffold something from here, what version of Holochain is it even compatible with?"

jost-s:
> This raises an interesting question -- should `hc-scaffold` even be available from a scaffolded flake? Seems to me that that would just create some sort of confusing recursive thing, where devs don't even realise that they should be asking themselves "if I scaffold something from here, what version of Holochain is it even compatible with?"

I have thought about that only briefly, and my initial reaction was that hc-scaffold isn't needed from a scaffolded project.

However, I imagine that a developer would expect to have all commands at their disposal inside of a Holonix shell, regardless of it being an ad-hoc shell or a project's shell definition. I don't expect them to be so familiar with Holonix and Nix to figure out which packages are included in which Nix expression. They'll know that hc scaffold is a Holochain-specific command and will try to run it from any Holonix shell, I suppose.

More importantly scaffolding isn't only for scaffolding new projects. It serves to configure existing projects graphically and add further components to it. It has to be available from a scaffolded project's flake too.

src/pages/quick-start/index.md
```diff
@@ -33,27 +33,28 @@ After installing Nix, close the terminal and open a new one.
...
-nix-shell (Nix) 2.13.2
+nix (Nix) 2.13.2
 ```
 
-Run the following command to set up the cache for precompiled Holochain executables:
+Run the following commands to set up the cache for precompiled Holochain executables:
 
 ```bash
-sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix --extra-experimental-features nix-command --extra-experimental-features flakes -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
+export NIX_CONFIG="extra-experimental-features = nix-command flakes"
```
jost-s, line 48: This doesn't work:
image

Enabling the experimental commands in the user's config bears the same result. So the command above needs to maintain the flags.

But instead of fixing this here, I'd rather move directly to the setup script.

jost-s (commented):
LGTM

As I said I suggest we skip this intermediate step and go straight to the setup script. I'm focusing on that now.

src/pages/quick-start/index.md
steveeJ: what if you change it to `--preserve-env=PATH,NIX_CONFIG`?

steveeJ: i think you also need to add another `--preserve-env` to make it effective, but i'm not sure. but if so, then `nix` should also work:

```
sudo -E --preserve-env=PATH,NIX_CONFIG nix run (...)
```

src/pages/quick-start/index.md
```diff
@@ -33,27 +33,28 @@ After installing Nix, close the terminal and open a new one.
...
+sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix  -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
```
steveeJ, line 49:
```suggestion
sudo -E --preserve-env="PATH,NIX_CONFIG" nix run nixpkgs/nixos-22.11#cachix -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
```

steveeJ:
the relevant changes of this PR are also in #372 thus this one can be closed

PR #365, opened 2023-02-10
update advanced installation instructions

steveeJ:

PR #367, opened 2023-02-10
advanced-install: remove history section

steveeJ:


steveeJ:
@jost-s i'm merging this since you suggested to remove this in #365 already and i didn't do it there.

PR #372, opened 2023-02-22
quick setup: introduce setup script

jost-s:
This PR replaces the installation instructions for Nix package manager and cachix setup with a setup script. An additional step in the script is now that Nix flakes and `nix` commands are enabled for the current user.

Secondly all `nix-shell` commands are updated to `nix develop` and `nix run` commands.


src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
 * 30GB+ available disk space
 * High Speed Internet connection
 
-## Install Holochain on Linux, macOS and Windows
+## Holochain on Linux, macOS and Windows
```
pdaoust, line 14: I'm curious about the reason for removing the word 'Install' here?

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
...
 !!! note Prerequisite for Windows: WSL2
 
-Holochain development uses the same tools across Mac, Windows, and Linux. However, the Nix toolkit, which we use to install and manage those tools, only works natively on Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
+The Nix package manager, which we use to install and manage Holochain development tools, only supports Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
 
 _Holochain is supported for the Ubuntu distribution under WSL2._
 !!!
 
-### Install the Nix Package Manager
+### Install the Nix Package Manager & set up Holochain binary cache
```
pdaoust, line 23:
```suggestion
### Install the Nix Package Manager and set up Holochain binary cache
```

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
...
-```
-
-## Scaffold Your First Holochain App
+### Scaffold Your First Holochain App
 
 Type the following at the command line:
 
 ```bash
-nix-shell https://holochain.love --run "hc scaffold example forum"
+nix run github:holochain/holochain#hc-scaffold -- example forum
```
pdaoust, line 36: This change is interesting, because it quietly delivers on something Art's been wanting for a while -- removing holochain.love. Seems like a good change to me 👍🏼

pdaoust (changes requested):
Looks great, and I'm so excited to see this happening. Only one stylistic change and one question about the removal of 'Install'.

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
...
 !!!
 
-### Install the Nix Package Manager
+### Install the Nix Package Manager & set up Holochain binary cache
 
 At a command line:
 
 ```bash
-sh <(curl -L https://nixos.org/nix/install) --daemon
+bash <(curl https://holochain.github.io/holochain/setup.sh)
```
steveeJ, line 28: does it really need to be bash? i don't think we're using exclusive bash features in the script.

```suggestion
sh <(curl https://holochain.github.io/holochain/setup.sh)
```

src/pages/get-building/index.md
```diff
@@ -14,7 +14,7 @@ For this example we will scaffold the canonical example, a To-do hApp.
 
 Run the holochain scaffolding tool by typing in your terminal:
 ```bash
-nix-shell https://holochain.love --run "hc scaffold web-app"
+nix run github:/holochain/holochain#hc-scaffold -- web-app"
```
steveeJ, line 17:
```suggestion
nix run github:/holochain/holochain#hc-scaffold -- web-app
```

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
 * 30GB+ available disk space
 * High Speed Internet connection
 
-## Install Holochain on Linux, macOS and Windows
+## Holochain on Linux, macOS and Windows
 
 !!! note Prerequisite for Windows: WSL2
 
-Holochain development uses the same tools across Mac, Windows, and Linux. However, the Nix toolkit, which we use to install and manage those tools, only works natively on Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
+The Nix package manager, which we use to install and manage Holochain development tools, only supports Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
```
steveeJ, line 18: why did you remove the "works natively"? i personally think that's a relevant information for windows users.

steveeJ (changes requested):
nice! almost there :blush:

src/pages/quick-start/index.md
jost-s: `sh` didn't work and I didn't want to spend more time on figuring it out. If you know how to make it work, welcome!

src/pages/quick-start/index.md
jost-s: I've found calling this "Installation" funny from the beginning. It's installing a package manager and then setting up a cache to get pre-compiled packages from. It's not "installing Holochain". After running these commands Holochain isn't installed, you can't run a Holochain command. It's unlike installing "Vue" or "npm".

It would be "Installing development environment tools for Holochain on Linux, macOS and Windows" I'd say, but that's a bit long for my taste.

src/pages/quick-start/index.md
jost-s: I find it doesn't add anything substantial and instead leaves room for interpretation. If I say "it's supported under Windows" or "it runs under Windows", it's implicit that it's a native binary for Windows.
When I say "it works under Windows, but not natively, you need to run it in a VM or an emulator or so" - well that's true for everything that's not supported under a certain OS. It's like saying "Visual Studio works under macOS, just not natively, you need to run it in a VM with Windows".

src/pages/quick-start/index.md
steveeJ: do you mean that the nix installation didn't for you either? that'd be very surprising for me because this is the way it's documented on nixos.org

jost-s: It does work when run as a command. It doesn't work inside of this script.

steveeJ: i'll let this one be a mystery as of now especially since we rely on bash inside the script now too

steveeJ (approved):
there now is a CI job testing this script on ubuntu and macos, and here's [a successful run](https://github.com/holochain/holochain/actions/runs/4290444975/jobs/7474468010) with [the most recent changes](https://github.com/holochain/holochain/pull/2017).

i tested it manually in a fresh ubuntu distribution of WSL2 and confirm it works :+1:

image

src/pages/quick-start/index.md
zippy, line 14:
```suggestion
## Install the Holochain Dev Environment
```

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
...
-```
-
-Run the following command to set up the cache for precompiled Holochain executables:
-
-```bash
-sudo --preserve-env=PATH $(which nix) run nixpkgs/nixos-22.11#cachix --extra-experimental-features nix-command --extra-experimental-features flakes -- use holochain-ci -m root-nixconf && sudo pkill nix-daemon
-```
-
-## Scaffold Your First Holochain App
+### Scaffold Your First Holochain App
```
zippy, line 31:
```suggestion
### Scaffold An Example Holochain App
```

zippy: This section was confusing in that it actually sounded like you were going to scaffold your OWN first holochain app, but the the script scaffolds OUR example hApp.  I think we need to re-think how we do this, but for now changing the title as above makes it a little clearer.

zippy (changes requested):
I've offered a few small changes.  I do think we should move forward with this as it's a great improvement.

src/pages/quick-start/index.md
```diff
@@ -11,49 +11,29 @@ pageStyleId: install
 * 30GB+ available disk space
 * High Speed Internet connection
 
-## Install Holochain on Linux, macOS and Windows
+## Install the Holochain Dev Environment
 
 !!! note Prerequisite for Windows: WSL2
 
-Holochain development uses the same tools across Mac, Windows, and Linux. However, the Nix toolkit, which we use to install and manage those tools, only works natively on Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
+The Nix package manager, which we use to install and manage Holochain development tools, only supports Mac and Linux. Linux can be run under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
```
jost-s, line 18:
```suggestion
The Nix package manager, which we use to install and manage Holochain development tools, only supports Mac and Linux. Please install Linux under Windows with [WSL2 (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install){target=_blank}.
```

PR #381, opened 2023-03-08
Mar 2, 2023 Adding page search

MatthewNichols:


pdaoust (approved):
Works beautifully! I didn't test it exhaustively, but it does what I'd expect/hope for it to do.

PR #390, opened 2023-04-28
switch to hello-world instead of forum

zippy:


pdaoust:
@zippy I've got both suggestions and a raft of tiny punctuation/formatting fixes; would you prefer I do the latter as a commit on `hello-world` rather than polluting the conversation with zillions of suggestions?

pdaoust:
This one is pretty stale now; the new getting started guide covers it.

PR #393, opened 2023-05-11
Add quantum time to Core Concepts and glossary

pdaoust:
I just discovered a new DNA modifier today, quantum time. @maackle @neonphog asking for your review because I'm not sure I even know what it is -- mostly just making a guess here! :D


src/pages/concepts/2_application_architecture.md
```diff
@@ -84,7 +84,15 @@ As a bundle of running code, a cell exposes the functions you write as its publi
...
+
+The origin time defines the earliest possible timestamp that data can have, or in other words the earliest date for which peers will try to create a window, while the quantum time defines the smallest window size. All windows are 'snapped' to a grid whose time axis starts at the origin time and whose units are a multiple of the quantum time.
+
+You generally don't need to worry about setting or changing values for the origin and quantum times in a DNA. The defaults set by the scaffolding tool are usually good enough. But if you find your network isn't synchronizing its data as efficiently as you like, it might be useful to play with these values.
```
maackle, line 94: Feel free to remove or modify if this is too much information

```suggestion
For the most efficient functioning of the network, the origin time should be set as close as possible to the time of the first action made by the first agent in the network. The scaffolding tool automatically sets the origin time to the time the DNA was created, which will suffice if you release your DNA soon after creating it, but if there is a long delay between creating and releasing your DNA, you should consider updating the origin time right before releasing. The network will function perfectly fine with an older origin time, but there will be some unnecessary extra overhead during gossip.

You generally don't need to worry about setting or changing the quantum time of a DNA. The default value works well for most networks. In some very exceptional cases, there may be some benefit to changing this value, which is why it's included as a DNA modifier, but in the vast majority of cases, it's best to leave it alone.
```

src/pages/references/glossary.md
```diff
@@ -725,6 +725,14 @@ See [public-key cryptography](#public-key-cryptography).
 
 To convert a [record](#record) into one or more [DHT operations](#dht-operation) and send them to the respective [validation authorities](#validation-authority) for [validation](#validation-rule), transformation into [record data](#record-data), and storage. This happens after it has passed the author's own copy of the [validation rules](#validation-rule). The validation authorities who are responsible for that entry's [address](#address) receive it, validate it, and if it's valid, store a copy of it and pass a [validation receipt](#validation-receipt) back to the author.
 
+#### Quantized gossip
```
maackle, line 728: I wonder if "quantized gossip" should even be in the glossary? I feel like it's mainly a term we used to distinguish the old gossip from the new "quantized" form of gossip. Like, it's synonymous with "new gossip". It just happens that the new gossip is characterized by being quantized. But we don't need to make special note of that fact, it's only special because it's the big change from the last version.

maackle (commented):
Some suggestions

src/pages/concepts/2_application_architecture.md
```diff
@@ -84,7 +84,15 @@ As a bundle of running code, a cell exposes the functions you write as its publi
...
+The DNA can also contain metadata: a name, description, **network seed**, **properties**, **origin time**, and **quantum time**. These can be changed either in a text editor or at installation time. When the network seed is changed, it **clones** a DNA, creating a new cell with identical functionality but an entirely separate history, network, and shared database. Properties can affect the new cell's runtime behavior (similar to configuration parameters or environment variables), and changes to properties also clone the DNA. Finally, the origin and quantum times affect the way that the network is going to exchange data.
```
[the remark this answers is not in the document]
pdaoust: hmm, I see what you mean. Wonder how I can phrase that.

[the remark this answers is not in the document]
pdaoust: I think I'll change it to 

```suggestion
The DNA can also contain metadata: a name, description, **network seed**, **properties**, **origin time**, and **quantum time**. These can be changed either in a text editor or at installation time. When the network seed is changed, it **clones** a DNA, creating a new cell with identical code but an entirely separate history, network, and shared database. Properties can affect the new cell's runtime behavior (similar to configuration parameters or environment variables), and changes to properties also cause the DNA to be cloned. Finally, the origin and quantum times affect the way that the network is going to exchange data.
```

src/pages/concepts/2_application_architecture.md
pdaoust: I really think this ought to be captured and shared somewhere. I have a struggle with giving TMI vs NEI (not enough info, new acronym -- you're welcome, world) in the Core Concepts and am never quite sure where to balance it. If there were a place that gave this sort of application note, I could just link out to it.

src/pages/references/glossary.md
pdaoust: On the one hand, I get what you're saying -- it's a term that only we care about internally. On the other hand, I'm not against giving TMI in the glossary; there's no danger of breaking a narrative flow here. The reason I put it in at all is that it serves as an 'anchor' to give context to the purpose of quantum time and origin time.

PR #397, opened 2023-08-10
Add cachix workaround to the quick start

ThetaSinner:


abe-njama:
@ThetaSinner Thanks for this.

Shouldn't the workaround come before the line,_"The Holochain Dev Environment is configured successfully on your system if you see a scaffolding version like..."_? I believe the error message comes up before the successful config of the dev environment.

ThetaSinner:
Probably not, I want people to try this section rather than skipping it because of a note up top. Then once the section gets into flow I'd rather not break out of it, because it's actually a continuous piece of writing as it stands. I think being a callout makes this visible enough at the bottom of the section to help people proceed if they do hit an error

jost-s:
@ThetaSinner I saw you didn't merge this. Feel free to merge PRs once they're approved just like with the Holochain repo.

ThetaSinner:
@jost-s Thank you, yes I meant to merge it 👍

PR #401, opened 2023-10-04
get-started: open links in new tabs

jost-s:
All links on the getting started page open in the current tab. I find that distracting and nowadays even unexpected. I suggest to open links in new tabs by default.


zippy (approved):

pdaoust:
This is wonderful and I agree with your rationale. That said, I wonder if we should consider this for the _whole_ website using some sort of 11ty plugin that transforms output. Could become a maintenance burden if we have to remember to put the external attr on each link.

pdaoust:
This one seems to be a drop-in! https://www.npmjs.com/package/eleventy-plugin-external-links?activeTab=readme

jost-s:
fixed by plugin #403

pdaoust:
I wrote a follow-up to this, #403 , which applies this to all links automatically. Okay to close this?

PR #403, opened 2023-10-05
Open non-relative links in new tab/window

pdaoust:
uses the plugin @aloskutov/eleventy-plugin-external-links


jost-s (approved):
Yeah, whoo!

PR #405, opened 2023-10-13
Holo docs for alpha

alastairong:


alastairong:
@pdaoust Really appreciate your work formatting, aligning styles, and getting the docs compiling! I was wondering about how to do that

pdaoust:
@alastairong no prob -- the build problem was apparently caused by a plugin called Prism, which highlights code samples at compilation time.

I've gone and done a big editing sweep; I've got some follow-up questions that I'll do as a review.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+#### Sign-up, sign-in, sign-out
+
+Users derive application-specific keys directly in Chaperone using their email and password. This is done directly in the iframe provided by the Chaperone connection manager, which the application developer can customize and must invoke somewhere in their user flow:
+
+```typescript
+client.signUp();
+client.signIn();
+```
+
+Since AdminWebsocket functionality is handled via Envoy, a service running on Holo hosts that handles the provisioning of hApps, Holo decides whether to show a "sign-up" or "log-in" form based on whether a cell needs to be provisioned for the user.
```
pdaoust, line 746: I don't quite understand this sentence -- is it saying that things normally accessible in the self-hosted context are not accessible in the Holo context because Envoy intercepts some of them? If that's the case, this is (or should be) true of the self-hosted context too -- admin websocket shouldn't be accessible to the UI. So the sentence on line 748 might also describe a call that UI devs shouldn't be making anyway.

(Check my facts; I might be wrong about this)

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+Since the host conductor is remote, UIs need to handle different connection states appropriately. Chaperone makes connection state information available to the client via events that the UI can listen for and handle. Example:
+
+```typescript
+client.on('agent-state', (agent_state: AgentState) => {
+  // Handle changes to agent state
+});
+```
+
+This generally results in a UI with a much more asynchronous programming pattern, whereas native Holochain UIs tend to be use a more synchronous programming pattern.
```
pdaoust, line 766: So far I've mostly seen native HC UIs programming in async style because it's all about waiting for WebSocket responses. Is there more beyond this that needs to be done -- some sort of mounting/unmounting of all zome calls based on agent state?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+    * All cells for a hApp are guaranteed to be co-located on the same Holo host.
+    * Cloned cells are treated as part of their parent hApp bundle and are thusly also co-located with the other cells in the hApp.
+* hApps are deployed onto the Holo network by an agent of the Cloud Console app, who is the designated manager for the hApp.
+* hApp managers are responsible for deploying application updates and paying host invoices.
+
+### Multi-tenant conductor
+
+Holo makes use of multi-tenant conductors to provide functionality typically expected of traditional web applications. Each conductor runs applications belonging to multiple users.
+
+By default, Holo will run read-only cells of deployed applications **under each host's agent** in addition to any user cells. These cells serve two purposes:
```
pdaoust, line 783:
```suggestion
By default, Holo will run **read-only cells** of deployed applications **under each host's agent** in addition to any user cells. These cells serve two purposes:
```

Thought that point deserves extra emphasis.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+#### Providing high-uptime nodes
+
+These nodes will send and receive gossip to sustain the network as long as hosts are online. They also ensure that when users connect to their host they already have the latest data available, since Holochain maintains a single local copy of a DHT for all cells of that DNA.
+
+#### Serving data to anonymous users
+
+For some use cases, it's not desirable to require users to create an account, for example, reading an article. Where developers make these use cases available to their users, the data is retrieved by the read-only instance.
+
+This behavior is available by default. Once a WebSDK client has been instantiated and before a user derives their agent keys, any zome calls will follow this pattern and have read access by default.
```
pdaoust, line 793: How does this work -- how does the stack know which functions are only going to need read access? What cap grant does an anonymous Chaperone connection use? AFAIU this would require an unrestricted grant on all zome functions.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+If you want to securely disable this functionality, you will need to implement a membrane proof with read-only restrictions. As a less secure alternative, you can:
```
pdaoust, line 795: * Which functionality specifically? the 'free pass to everything' functionality?
* I'm confused by the phrase 'implement a membrane proof with read-only restrictions' -- do you mean a membrane proof validation function that allows the read-only agent's membrane proof paired with a general validation function that denies any source chain writes? Or are source chain writes denied at the conductor level somehow?
* I'm not sure we need to give a less secure alternative -- I'd prefer just giving one (secure) piece of advice to everyone. Thoughts?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+1. Set up your UI lifecycle to call the sign-up/sign-in form and ensure credentials are set up before requesting any data from the DHT, and
+2. Ensure that `.signUp({cancellable: false})` or `.signIn({cancellable: false})` are configured.
+
+#### Admin API WebSocket access
+
+Due to the security implications of multi-tenant conductors, `AdminWebsocket` (and some `AppWebsocket`) functionality is not directly exposed to Holo clients. Instead, this functionality is exposed via Holo WebSDK methods where appropriate.
```
pdaoust, line 802:
```suggestion
Due to the security implications of multi-tenant conductors, `AdminWebsocket` (and some `AppWebsocket`) functionality is not directly exposed to Holo clients. Instead, this functionality is exposed indirectly via Holo WebSDK methods where appropriate.
```

Am I right about that?

But also, as mentioned above, this also should be true of native HC UIs. The fact that it isn't yet is just because the restriction isn't enforced yet.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+### Membrane proofs
+
+Holo strongly recommends membrane proof implementations for production or long-running applications. However, as noted earlier, Holo installs and runs cells of deployed applications under each host’s agent to provide high-uptime DHT nodes and to serve data to anonymous users and Web2 participants who do not have keys.
```
pdaoust, line 806:
```suggestion
Holo strongly recommends membrane proof implementations for production or long-running applications. However, as noted earlier, Holo installs and runs read-only cells of deployed applications under each host’s agent to provide high-uptime DHT nodes and to serve data to anonymous users and Web2 participants who do not have keys. **These cells always have a zero-byte membrane proof.**
```

I've got this right, right? This feels like the important take-home message; if you want to create a restricted membership but still allow the read-only instances to serve public content and maintain DHT resilience, you've got to look for the read-only membrane proof.

(I have separate concerns about the easy forgeability of a zero-byte membrane proof that would allow a non-host access to any DHT they wanted, if the following were implemented in a DNA -- am I wrong about how that works?)

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+Membrane proofs that restrict this read-only access prevent Holo from providing such affordances for your hApp. This may be an acceptable trade-off or desired behaviour. However, If you want to benefit from always-on nodes and read-only access while having a membrane proof, we you will need to configure your membrane proof logic to allow "unpermissioned" read-only access like below:
```
pdaoust, line 808:
```suggestion
DNAs that require membrane proofs restrict this read-only access and thus prevent Holo from providing such affordances for your hApp. This may be an acceptable trade-off or desired behaviour. However, If you want to benefit from always-on nodes and read-only access while having a membrane proof, you will need to configure your membrane proof logic to allow "unpermissioned" read-only access like below:
```

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+```rust
+pub fn is_read_only_proof(mem_proof: &MembraneProof) -> bool {
+    let b = mem_proof.bytes();
+    b == &[0]
+}
+
+pub fn validate_joining_code(
```
pdaoust, line 816: This function might be misconstrued as an official callback. I wonder if we could demonstrate the use of this function within `validate()`?

It also seems like it needs validation code to deny all source chain writes past the genesis records -- or is this enforced at the conductor level?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+    Ok(InitCallbackResult::Pass)
+}
+```
+
+In general, you should not list functions in the above code that _write_ to the source chain, only that read from the source chain.
+
+Finally, if you want to make sure that anonymous agents can't write to their own source chain, you should add the following protection to any function that writes to the source chain:
+
+```rust
```
pdaoust, line 862: As mentioned above, I'd recommend moving this restriction to the integrity zome's `validate()` function, so that any records past `InitComplete` are prevented at both the conductor and DHT level and can't be circumvented by simply hot-swapping coordinator zomes. I don't know what this would look like -- how many different types of `Op` you'd have to put this guard on. Probably all of them 🤷🏼‍♂️ The one problem with this is that you'd have to do it for _each_ integrity zome, and some of those zomes might be third-party.

pdaoust (changes requested):

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+## Get started
+
+Holo hosting is for Holochain applications. It is assumed that you have familiarity with Holochain.
+
+https://developer.holochain.org/get-started/
+
+There are two pathways in this section --- scaffolding a new hApp or migrating an existing hApp. Both pathways contain largely the same information, so choose the pathway that best applies to your situation.
+
+In both cases there are no required DNA changes for Holo Hosting --- only UI changes. However, Holo operates under a different context and we highly recommend that you read the core concepts, in particular the sections on membrane proofs and anonymous access, as these **will need DNA integrity zome changes to enable certain features**.
```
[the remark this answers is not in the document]
alastairong: Ah! I meant the Holo Core Concepts which is the section below that you're referring to. I wonder how we structure this better

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+npm install
+```
+
+### UI
+
+The biggest difference between a Holo and pure Holochain application is in the UI. The scaffolding tool should automatically have generated everything required here, so we'll simply point out the primary differences.
+
+In a Holo setting we use the [`@holo-host/web-sdk`](https://www.npmjs.com/package/@holo-host/web-sdk) library instead of `@holochain/client`. WebSDK provides an instance of `AppAgentWebsocket`, which is almost the same as `AppWebsocket`.
+
+```typescript
```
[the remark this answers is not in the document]
alastairong: It should be in App.vue. How should we add it in? Do we just spell it out like "Let's look at App.vue"? Or is there a smarter way?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+  components: {
+    //...
+  },
+  data(): {
+    client: AppAgentClient | undefined; // Ensure that client is not AppAgentWebsocket
+    //...
+  } {
+```
+
+You can then initialize a context-dependent client, assuming you want your UI to be compatible with both Holo and Holochain.
```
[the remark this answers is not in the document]
alastairong: The data() function itself doesn't check right?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+    IS_HOLO: boolean;
+  } {
+    return {
+      client: undefined,
+      loading: true,
+      IS_HOLO: ['true', '1', 't'].includes(import.meta.env.VITE_APP_IS_HOLO?.toLowerCase()), // You should set this as an env variable
+    };
+  },
+
+  async mounted() {
```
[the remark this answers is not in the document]
alastairong: mounted()? This is a Vue lifecycle function to specify logic that occurs when the UI elements are mounted to the DOM. And yes we add logic in here specific to Holo.

I haven't had all the experience of writing documentation that you have so I'm not sure what we really need to call out about this - seemed self-apparent to me. Can you suggest something?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+          appName: 'forum-app', // Display name on the credentials form. You can also set it in Cloud Console when deploying
+        }
+      });
+
+      this.client = client
+    } else {
+      this.client = await AppAgentWebsocket.connect('', 'vue-default');
+      this.loading = false;
+    }
+```
```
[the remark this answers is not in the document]
alastairong: Yes. I didn't think it was important but you're right that it's literally the only thing left out

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+        authFormCustomization: {
+          appName: 'forum-app', // Display name on the credentials form. You can also set it in Cloud Console when deploying
+        }
+      });
+
+      this.client = client
+    } else {
+      this.client = await AppAgentWebsocket.connect('', 'vue-default');
+      this.loading = false;
+    }
```
alastairong, line 465:
```suggestion
    }
}
```

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+  if (this.IS_HOLO) {
+  // ...
+
+    // register event handler for agent-state.
+    client.on('agent-state', (agent_state: AgentState) => {
+      this.loading = !agent_state.isAvailable
+    });
+
+    // ...
+```
```
alastairong, line 485:
```suggestion
    }
}
```
```

File: src/pages/references/enable-holo.md [hunk 18]
```diff
@@ -0,0 +1,896 @@
...
+
+Now your UI can connect to the Holo network and read forum posts. However, users can't post or comment yet because they don't yet have a source chain. They are considered "anonymous" and can only request public data from the network.
+
+!!! note
+Unlike in pure Holochain, you cannot assume that a user has provided their desired keys when the connection is established. This may require changes to your application logic. You can read more at https://link.to/keys/documentation
```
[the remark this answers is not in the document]
alastairong: Yes they may not have generated their keys yet. When they haven't, we have some throwaway / temporary keys generated for them. To be honest I'm unclear as to why.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+Now you're ready to build and test!
+
+### Building and testing
+
+Holo provides the `holo-dev-server` binary, which simulates the Holo network locally for development. `holo-dev-server` serves a copy of Chaperone, the JavaScript library that connects the browser to the Holo network, and runs a Holochain conductor. Like the real Holo network, `holo-dev-server` uses `.happ` bundles, which do not include a UI. (The `hc` developer tool creates both a `.happ` bundle and a `.webhapp` bundle for you.)
+
+To include `holo-dev-server` in your development environment, update your project's `flake.nix` file to the below, and re-enter the shell:
+
+```nix
```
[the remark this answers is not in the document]
alastairong: We can't as Holo Network itself is on 0.2.x

I'm happy to keep this as a feature branch until we get to 0.2.x stable

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+    "start:holo": "AGENTS=2 npm run network:holo",
+    "network:holo": "npm run build:happ && UI_PORT=8888 concurrently \"npm run launch:holo-dev-server\" \"holochain-playground ws://localhost:4444\" \"concurrently-repeat 'VITE_APP_CHAPERONE_URL=http://localhost:24274 VITE_APP_IS_HOLO=true npm start -w ui' $AGENTS\"",
+    "launch:holo-dev-server": "sudo ./holo-dev-server workdir/test-app.happ",
+```
+
+These scripts assume that you have a `start` script setup in a `ui` workspace, which you should have if you created the hApp using our scaffolding tool.
+
+You now have a fully functional Holo app up and running!
+
+!!! info
```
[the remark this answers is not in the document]
alastairong: Do you mean the chaperoneUrl advice? I'm sure there is. We can add it to the todo list

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+This process assumes you already have a Cloud Console account. If you do not, you can sign up at https://register.holo.host
+
+### Deploy UI
+
+Holo does not currently offer UI hosting. hApp managers will need to deploy their UI separately for now.
+
+!!! info
+Holo supports "headless hosting" where there is no Holo-enabled UI. In this case Holo will maintain copies of the DHT, but will not hold any source chains. Holo hosting requires <membrane proof recommendation?>
+
+Holo does not support direct programmatic/API access to deployed hApps.
```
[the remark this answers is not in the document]
alastairong: It's just not productised yet. We actually have a POC implementation of this in our HRL gateway that fractal-tribute uses.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+!!!
+
+### Deploy hApp to Holo network
+
+1. Log into Cloud Console.
+2. Click on "Add a hApp".
+3. Fill out hApp details:
+    * "Link to hApp file" must point to a publicly accessible `.hApp` file.
+    * (Optional) "URL of hosted UI" is the URL which your users will access your UI from.
+    * (Optional) Configuration of memproof/registration code.
```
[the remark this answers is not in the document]
alastairong: No this is not a deploy guide and the content re: how to properly configure registration code / memproof set up is part of a deploy guide. We might revisit this if there is a learning pathway issue but right now I'd rather we prepare a "deploy guide" as a standalone.

src/pages/references/enable-holo.md
alastairong: Should it be true of native HC UIs? Or do you mean "exposed to the end user". Because I mean that someone developing a UI for Holohosting using web-sdk cannot make AdminWebsocket calls.

src/pages/references/enable-holo.md
alastairong: That is correct. I'm wondering if we should clarify that "Holo CAN install..." i.e. you can disable this functionality by simply not providing the zero-byte membrane proof. What do you think?

Re: the easy forgeability, yes that is also something on our radar that we don't have a good answer to currently. Long term the solution is to allow HoloPorts to request membrane proofs from the happ manager's infrastructure directly using some sort of secret.

src/pages/references/enable-holo.md
alastairong: @robbiecarlton Can you answer this question?

src/pages/references/enable-holo.md
alastairong: @robbiecarlton This one too

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+### Testing
+
+Holo provides the `holo-dev-server` binary, which simulates the Holo network locally for development. `holo-dev-server` serves a copy of Chaperone, the JavaScript library that connects the browser to the Holo network, and runs a Holochain conductor. Like the real Holo network, `holo-dev-server` uses `.happ` bundles, which do not include a UI. (The `hc` developer tool creates both a `.happ` bundle and a `.webhapp` bundle for you.) When you scaffold an app with the `--holo` flag, its dev environment will provide `holo-dev-server` on the command line for you.
+
+1. Ensure you are in the nix development environment by checking the shell prompt. Run `nix develop` in the root of your project's folder if you aren't.
+2. Run `npm start:holo`, which:
+    1. Starts `holo-dev-server' and provisions instances (cells) of your DNA for two agents,
+    2. Automatically opens an instance of a network inspector called [Holochain Playground](https://github.com/darksoil-studio/holochain-playground) in your browser, and
+    3. Runs a web server to serve the UI.
```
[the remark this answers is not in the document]
alastairong: Yea I guess technically. It's an out-of-the-box thing for most UI frameworks to make the UI available at localhost right?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+3. Access your two agents on your browser at `http://localhost:8888` and `http://localhost:8889`.
```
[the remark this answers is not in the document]
alastairong: At launch, they are two separate anonymous instances with throwaway keys. Depending on application logic they might just immediately direct to a login/signup page, or allow read-only use at first.

I don't know exactly how we keep their chaperone instances separate.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+Save the file, and you should see that the windows have auto-updated to show something that should look like this:
+
+![Two web browser windows showing basic widgets to manipulate the shared to-do list](assets/img/scaffolded_todos.png)
+
+You now have a fully functional Holo app up and running!
+
+## Migrate from a pure Holochain app
+
+Only minor UI changes are technically required for Holo hosting. However, Holo operates under a different set of assumptions and we highly recommend that you read the [Core Concepts](/concepts/). It is likely that you will want to reconsider certain UX flows.
```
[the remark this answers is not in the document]
alastairong: Yup - let's agree how we want to structure the Holo "core concepts" and what name we even want to call it, given the confusion with Holochain Core Concepts

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+      this.client = await AppAgentWebsocket.connect('', 'vue-default');
+      this.loading = false;
+    }
+```
+
+The WebSDK client loads and connects to an iframe called Chaperone. Chaperone is a key and connection manager to the Holo network and is invisible outside a few specific cases.
+
+Because the Holo network acts like a remote conductor, you'll need to handle connectivity issues. Changes to the connection are emitted and you can register an event handler if the client is an instance of WebSDK. You can see the full list of emitted events here: https://github.com/Holo-Host/web-sdk
+
+```typescript
```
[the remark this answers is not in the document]
alastairong: Yea...structurally it's part of the snippet above actually. I manually split it out but perhaps that's a bad idea that causes confusion. I'll make a separate commit to address this.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+!!! note
+Unlike in pure Holochain, you cannot assume that a user has provided their desired keys when the connection is established. This may require changes to your application logic. You can read more at https://link.to/keys/documentation
+!!!
+
+Holo uses "sign-up" and "sign-in" terminology but there is no external authentication process. In both processes a user (re)derives their keys from email and password, with sign-up also triggering the installation of new hApp cells on a hosting device in the Holo network.
+
+So let's add sign-up and sign-in functionality to the app.
+
+```html
```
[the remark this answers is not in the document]
alastairong: It's still part of App.vue in the scaffold I think? Just in the html section.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+  <div id="content">
+    <h2>All Posts</h2>
+    <AllPosts></AllPosts>
+    <span style="margin-bottom: 16px"></span>
+    <CreatePost></CreatePost>
+  </div>
+</main>
+```
+
+```typescript
```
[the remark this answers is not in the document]
alastairong: App.vue again. Lower down after the data() section there's a methods section.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+  // ...
+}
+```
+
+These trigger the display of a full screen credentials modal from the Chaperone iframe, defaulting to either a sign-up or sign-in form. The user can also toggle between them directly.
+
+![](https://hackmd.io/_uploads/r1WJbrkC3.png)
+
+During sign-up a registration code field can be shown to enable [**membrane proof**](/glossary/#membrane-proof) workflows. Holo strongly recommends that you implement a membrane proof for production or long-running applications. However, membrane proofs that restrict read-only access prevent Holo from providing always-on nodes and [read-only access to applications](https://link.tbd).
```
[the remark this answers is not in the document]
alastairong: Thanks. Yes it's good to keep the info aligned across both.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+          shellHook = ''
+            nix-env -f "https://hydra.holo.host/channel/custom/holo-nixpkgs/${holoDevServerChannel}/holo-nixpkgs/nixexprs.tar.xz" -iA holo-dev-server
+          '';
+        };
+      };
+    };
+}
+```
+
+To connect your application to `holo-dev-server`, point the client's `chaperoneUrl` to `https://localhost:24274`.
```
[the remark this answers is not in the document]
alastairong: Ok any action required here? If not please hit resolve :)

src/pages/references/enable-holo.md
alastairong: AdminWebsocket calls aren't exposed to the UI?

src/pages/references/enable-holo.md
alastairong: So my impression is that in native Holochain you have serial async approach

I.e. 
```
async function foo() -> Promise<Bar> { ... }
const result = await foo

```

In Holo UIs you tend to have more of a listener pattern i.e.

```
async function foo() -> Promise<Void> { ... }

listener.on('event_that_foo_triggers') => { ... }
```

src/pages/references/enable-holo.md
alastairong: Thanks. This is old. I will update in a separate PR

src/pages/references/enable-holo.md
alastairong: Happy to remove the less secure alternative. It's kinda hacky.

alastairong: I think generally I need to re-write this section

src/pages/references/enable-holo.md
robbiecarlton: Paul's right, this would be more correct.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,896 @@
...
+
+::: output-block
+```text
+Collection "my_todos" scaffolded!
+```
+:::
+
+Holo strongly recommends that you implement a [**membrane proof**](/glossary/#membrane-proof) for production or long-running applications. However, membrane proofs that restrict read-only access prevent Holo from providing always-on nodes and [read-only access to applications](https://link.tbd).
+
+If you want to benefit from always-on nodes and read-only access while having membrane proofs, you will need to implement special "Holo-safe" logic. You can [read more here](https://link.to.core.concepts).
```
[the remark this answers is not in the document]
pdaoust: I'd like to have a Holo core concepts section, sure. Then we could link out to it from all sorts of places.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: ohhhhhhhhhhh it's the UI framework itself that has a local dev server? got it.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: if the function already exists in the Vue file (say, as a stub) let's just be explicit -- "edit the two Vue lifecycle functions so they read as follows" or something like that.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: okay, so it sounds like all the edits are to `App.vue` then? Let's put something near the top to say "all edits you'll be making will be to your UI's entry point -- in this case, `App.vue`" and then give a bit of context for each snippet (e.g., "put this stuff in the section that starts with ___")

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: aw dang! Good to know. Yes, let's keep this as a feature branch unless we get a pull (from you, Rob, some business decision-maker, whoever) to get that documentation online. We'll want to put a caveat at the top that hApp devs will need to either start scaffolding from some 0.2 channel of Holonix or upgrade their DNAs to 0.2.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: I have no idea why I've got it as a suggestion that changes nothing 🤔 I'll think of a way to elaborate on what's needed.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
pdaoust: yes, the chaperoneUrl advice, along with `VITE_APP_IS_HOLO` env var. So you've got three scenarios:

* `build` w/o any env vars builds the UI for native HC
* `build:holo` passes `VITE_APP_IS_HOLO=true` and a production `VITE_APP_CHAPERONE_URL` env vars (with alternatives for other front-end toolkits, of course)
* for local dev, allow the parent `package.json` to specify those env vars, as in the example you gave.

src/pages/references/enable-holo.md
pdaoust: ohh, got it, so you wouldn't use `await` as much in Holo-hosted? Is that because of higher latency / greater risk of disconnection?

src/pages/references/enable-holo.md
pdaoust: How to make this performant is a question though -- you don't want to be doing a `must_get_agent_activity` on every single zome call. I suppose it could work like this when validating:

1. If the preceding source chain action is an agent key, require a non-read-only membrane proof preceding it.
2. If the preceding source chain action is anything else and a validator has marked it valid, it's assumed it was also subjected to these two rules, so the action currently being valid must also be writable.

This recursive rule inductively validates a whole chain but only requires a single DHT get per validation. (Of course, it also depends on the honesty of validators.)

src/pages/references/enable-holo.md
pdaoust: annoyingly, the way GH's suggestion fences work, the closing fence is lost. I ran into this yesterday -- had to edit it locally and push the changes.

src/pages/references/enable-holo.md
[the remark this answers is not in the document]
alastairong: Ah I see. Yea

src/pages/references/enable-holo.md
pdaoust: Another problem (mentioned in chat, but noting here for visibility) is that this would have to be explicitly coded into the validation function of each integrity zome, which would prevent you from using third-party zomes. All in all, the validation approach is 'more correct' (= secure) if zero-byte memproofs are to be used, but very challenging to make workable in our current composability model -- Integrity zomes are closed to both modification and extension.

src/pages/references/enable-holo.md
pdaoust: I looked at the actual current state and it looks like you resolved this 👍🏼

src/pages/references/enable-holo.md
```diff
@@ -448,16 +449,27 @@ You can then initialize a context-dependent client, assuming you want your UI to
       IS_HOLO: ['true', '1', 't'].includes(import.meta.env.VITE_APP_IS_HOLO?.toLowerCase()), // This will be passed later as an env variable
     };
   },
+```
+
+The WebSDK client loads and connects to an iframe called Chaperone. Chaperone is a key and connection manager to the Holo network and is invisible outside a few specific cases. Let's initialise this client in the Vue `mounted()` lifecycle hook.
```
pdaoust, line 454:
```suggestion
The WebSDK client loads and connects to an iframe called Chaperone. Chaperone is a key and connection manager to the Holo network and is invisible outside a few specific cases. Let's initialize this client in the Vue `mounted()` lifecycle hook.
```

I am so sorry; we've standardised on American spelling. It pains me too. (my spell checker tells me 'standardised' is spelled wrong and I'm stubbornly ignoring it. But I would write 'standardized' in the dev docs)

pdaoust (approved):

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+
+The most secure way to restrict access to a hApp's data is with [**membrane proofs**](/references/glossary/#membrane-proof). A membrane proof is supplied at installation time and acts like an ID card, allowing the hApp to admit valid members only. However, this also prevents Holo hosting devices from provisioning read-only cells.
+
+This is because read-only cells always use a zero-byte membrane proof. This is essentially the same as no membrane proof, so this is only appropriate for hApps that contain data that should be publicly readable but not anonymously writable. (Note that, in the future, the production Holo hosting network may permit you to implement logic that provisions membrane proofs to read-only nodes, allowing you to use Holo for high-uptime while preventing anonymous read access, but this work is currently out of scope.)
+
+To implement logic that allows read-only cells to join a DNA's network but not write data, your integrity zome's validation function should check the author's membrane proof, then allow them to join the network but reject CRUD operations if the proof is zero bytes.
```
alastairong, line 807:
```suggestion
In order to provide high-uptime nodes and anonymous hosted user access via read-only cells while keeping your hApp safe from defacement, you'll need to restrict what a read-only cell can do. A read-only cell is **read-only merely by convention**; that is, because you're checking for read-only cells and handling writes differently.

The most secure way to restrict access to a hApp's data is with [**membrane proofs**](/references/glossary/#membrane-proof). A membrane proof is supplied at installation time and acts like an ID card, allowing the hApp to admit valid members only. However, this prevents Holo hosting devices from provisioning read-only cells unless explicit affordances are made.

The Holo Network expects and uses a zero-byte membrane proof for read-only cells. A publicly known membrane proof without restriction is essentially the same as no membrane proof. To implement logic to allow read-only cells to join a DNA's network but not write data, your integrity zome's validation function should check the author's membrane proof, then allow them to join the network but reject CRUD operations if the proof is zero bytes.

(Note that, in the future, the production Holo hosting network may permit you to implement logic that provisions membrane proofs to read-only nodes, allowing you to use Holo for high-uptime while preventing anonymous read access.)
```

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+
+The following code shows an example of how to do both of these things. This isn't necessarily the most performant solution, as it requires a validator to retrieve the author's entire source chain (excluding entry data). A better-performing solution might rely on [inductive validation](/references/glossary/#inductive-validation) of the author's source chain.
```
alastairong, line 809: Why can't we do the better-performing one? If people are basically going to copy and paste this stuff we should write something we are happy with. Especially since changing this will be an integrity zome change.

src/pages/references/enable-holo.md
pdaoust: brilliant, thanks!

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+        .find(|record| match record.action.hashed.content {
+            Action::AgentValidationPkg(_) => true,
+            _ => false
+        })
+        .map(|record| match &record.action.hashed.content {
+            Action::AgentValidationPkg(action) => action.membrane_proof.clone(),
+            _ => None
+        });
+
+    Ok(!is_read_only_membrane_proof(membrane_proof.unwrap()))
```
zo-el, line 867: The unwrap in this statement can panic because the agent could have no mem-proof, right?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+            // Here the membrane proof itself is being validated.
+            // The AgentValidationPkg action produces two different ops that
+            // contain a membrane proof; let's check both of them.
+            if let Action::AgentValidationPkg(action) = match &op {
+                Op::StoreRecord(store_record) => Some(store_record.record.signed_action.hashed.content.clone()),
+                Op::RegisterAgentActivity(register_agent_activity) => Some(register_agent_activity.action.hashed.content.clone()),
+                _ => None
+            }.unwrap() {
+                return  validate_membrane_proof(action.membrane_proof.clone());
+            }
```
zo-el, line 890: This if statement needs an else

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+        },
+
+        hdi::prelude::ActionType::Create | hdi::prelude::ActionType::Update | hdi::prelude::ActionType::Delete => {
+            // If not, deny writes to read-only cells.
+            match op.entry_data().unwrap() {
```
zo-el, line 895: The unwrap can panic too right? because the initial creates do not have entries just actions

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+    let mut fn = BTreeSet::new();
+    fns.insert((zome_info()?.name, "get_article".into()));
+    fns.insert((zome_info()?.name, "get_all_articles".into()));
+
+    let functions = GrantedFunctions::Listed(fns);
+    create_cap_grant(CapGrantEntry {
+        tag: "".into(),
+        access: CapAccess::Unrestricted,
+        functions,
+    })?;
```
zo-el, line 956: This create is going to require write access, and with the above validation rule, it will block the read-only mem-proof from committing this grant right?

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,970 @@
...
+
+```rust
+pub fn set_cap_tokens() -> ExternResult<()> {
+    let mut fn = BTreeSet::new();
+    fns.insert((zome_info()?.name, "get_article".into()));
+    fns.insert((zome_info()?.name, "get_all_articles".into()));
+
+    let functions = GrantedFunctions::Listed(fns);
+    create_cap_grant(CapGrantEntry {
+        tag: "".into(),
```
zo-el, line 953:
```suggestion
        tag: "read-only".into(),
```

src/pages/references/enable-holo.md
pdaoust: This is just to get around the awkward fact that I've already checked the action type so I know the op's associated action is `AgentValidationPkg`. All I'm doing here is `if let` matching on a condition I know must be true. I've simplified it in a later commit.

src/pages/references/enable-holo.md
pdaoust: Hm, yeah, looks like `Action::entry_data()` is guaranteed to return something meaningful for new-entry actions (even for private entries, and even for ops that don't carry entry bytes), but not for delete-entry actions. Nice catch, thank you.

src/pages/references/enable-holo.md
pdaoust: nice idea.

src/pages/references/enable-holo.md
pdaoust: done, except for review on the code of the better-performing one.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,960 @@
...
+#[hdk_extern]
+pub fn validate(op: Op) -> ExternResult<ValidateCallbackResult> {
+    // Here the membrane proof itself is being validated.
+    // The AgentValidationPkg action produces two different ops that
+    // contain a membrane proof; let's check both of them.
+    if let Action::AgentValidationPkg(action) = match &op {
+        Op::StoreRecord(store_record) => Some(store_record.record.signed_action.hashed.content.clone()),
+        Op::RegisterAgentActivity(register_agent_activity) => Some(register_agent_activity.action.hashed.content.clone()),
+        _ => None
+    }.unwrap() {
```
pdaoust, line 916: crap, this will regularly panic now, I bet. Just about to push a fix.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,960 @@
...
+    // update) special system entries:
+    //
+    // * `AgentPubKey`
+    // * `CapGrant`
+    // * `CapClaim` (optional)
+    //
+    // When allowing cap grants, we're unable to check whether the grant should
+    // be allowed -- that has to happen in the coordinator zome.
+    if let ActionType::Create = action_type {
+        if let EntryType::AgentPubKey | EntryType::CapClaim | EntryType::CapGrant = op.entry_data().unwrap().1 {
```
zo-el, line 855: I am always sceptical of using unwrap, especially in the integrity zome.
Would it be better to handle the unwrap?

PR #406, opened 2023-10-24
replace Prism with highlight.js

pdaoust:
The syntax highlighting in 11ty's default plugin, which uses Prism, is a bit limited -- it doesn't have any language definitions that allow for nested languages (e.g., JS in HTML). This PR replaces it with highlight.js, another popular highlighting plugin.


MatthewNichols (approved):
This is great. I suggest adding an outline of the way that Syntax highlighting works to the README since it doesn't use the 11ty plugin that a lot of people use (14k+ weekly downloads on npm).  But no reason to wait on deploy.

PR #408, opened 2023-12-14
revert accidental merge of draft Holo getting started guide

pdaoust:
Somehow -- I don't know how -- I managed to locally merge the `holo-docs-for-alpha` branch into `main` and then push to GitHub, thus closing PR #405 . This reverts that disaster.

PR #409, opened 2023-12-15
re-add Holo getting started guide

pdaoust:
@alastairong @zo-el @marycamacho I somehow managed to accidentally merge `holo-docs-for-alpha` into `main` and then push it to GitHub on Tuesday, triggering a premature deploy of the guide. #408 fixes that, but it means I have to create a new branch and PR for the Holo getting started guide. So... here it is. Let further conversations and QA ensue from this point.


src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,1012 @@
...
+                    return Ok(true);
+                }
+
+                // If, however, we have seen the `InitZomesComplete` action,
+                // the op being validated happened after `init` but the prior
+                // action we're currently inspecting happened before. Skip this
+                // and all other `init`-related CRUD actions, and keep looking
+                // until we find the membrane proof.
+            }
+            (Action::Create(_), Some(EntryType::App(_))) | (Action::Update(_), Some(EntryType::App(_))) | (Action::Delete(_), _) => {
```
robbiecarlton, line 918: are these cases separated just for tidiness? nbd but I think I prefer them all as one match case, to make it clear that we handle all these cases the same way.

robbiecarlton (approved):
Couple of non blocking suggestions

src/pages/references/enable-holo.md
pdaoust: I don't remember the reason! Maybe for clarity? Makes sense to collapse them.

src/pages/references/enable-holo.md
```diff
@@ -0,0 +1,1012 @@
...
+
+::: output-block
+```text
+Collection "my_todos" scaffolded!
+```
+:::
+
+### Testing
+
+Holo provides the `holo-dev-server` binary, which simulates the Holo network locally for development. `holo-dev-server` serves a copy of Chaperone, the JavaScript library that connects the browser to the Holo network and manages user keys, and runs a Holochain conductor. Like the real Holo network, `holo-dev-server` uses `.happ` bundles, which do not include a UI. (The `hc` developer tool creates both a `.happ` bundle and a `.webhapp` bundle in the `workdir/` folder for you.) When you scaffold an app with the `--holo` flag, its dev environment will provide `holo-dev-server` on the command line for you.
```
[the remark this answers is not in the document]
pdaoust: thanks; I might've misunderstood what Chaperone is then. My understanding is that it's a JS file that:

* lives in an iframe
* communicates with the UI by via COMB on the Chaperone side and the web-sdk on the UI side
* handles signup/login flow
* asks the matchmaker for hosts to connect to
* connects to a host

pdaoust:
Closing cuz Holo is pivoting.

PR #412, opened 2024-01-08
Feat/mermaid rendering

mattyg:
**Overview**
- add client dependency mermaid
- render mermaid charts in code blocks when language is set to `mermaid`
- bundle mermaid dep and mermaid client code in separate bundle

Mermaid graphs can now be included with the following markdown:

    ```mermaid
    graph TB
       A --> B
       B --> C
    ```

Unfortunately the mermaid dependency is quite large ~3MB. The alternative would be to build mermaid graphs to SVGs and include those, but that would mean either more work to setup tooling to automate this, or else more friction for content editors to generate graphs. IMO for now its a good start.


pdaoust:
I'd love to render all the mermaid diagrams to SVG on the server using something like puppeteer, but yeah, it would require some extra steps in CI. And it'd increase build time/cost.

I'm not so worried about the friction for content editors; they can just run `npm build` or `npm dev` and see what it looks like -- or get Netlify to generate a build.

But for now, this makes Mermaid diagrams possible, and that's great! A couple questions and requests:

* Could you document what the scripts now do in [README.md](https://github.com/holochain/docs-pages/blob/main/README.md?plain=1#L96)? I don't know what the consequences of letting rollup handle TS compilation and bundling, and it'd be good to know e.g., what you should add to where if you wanted to create another bundle.
* Could you add one non-accessible Markdown page that lets us test Mermaid in this PR?

mattyg:
> I'm not so worried about the friction for content editors; they can just run `npm build` or `npm dev` and see what it looks like -- or get Netlify to generate a build.

We would need some tooling for it to work this seamlessly (i.e. parsing all the markdown for mermaid strings, running the mermaid cli to render them as svg, copying the svg into assets, replacing the mermaid content with the svg path). It would be a pain to do manually.

> Could you document what the scripts now do in [README.md](https://github.com/holochain/docs-pages/blob/main/README.md?plain=1#L96)? 

done

> I don't know what the consequences of letting rollup handle TS compilation and bundling, and it'd be good to know e.g., what you should add to where if you wanted to create another bundle.

It doesn't really make a difference, it just removes the step in the npm scripts to run `tsc`, as rollup handles it. I did add minifying to the rollup steps as well using Terser https://terser.org

To create another bundle you would just add it to rollup.config.js, which exports a list of bundles to generate.

mattyg:
Also just a heads up: bundling and minifying mermaid is quite slow (~25s for me). The alternative would be to include an already bundled mermaid js file, but the downside would be we couldn't manage the version via package.json.

pdaoust:
As a first step toward getting back on the rails here, I'd like to merge this PR. Some thoughts:

* My colleague is going to run it by [Lighthouse](https://lighthouse-metrics.com/) to see what they think of the increased page weight.
* Is it possible to bundle just the features we need? For me the most desperately needed one is sequence diagrams, but I can also see value in flowcharts and block diagrams. There is absolutely zero information about this on their docs site.
* Maybe each page could opt into including Mermaid, so you wouldn't have to actually load the asset until it was time to see some diagrams?

mattyg:
Unfortunately mermaid is very heavy (3MB!), beyond what is reasonable to add to the bundle.

Upon discussion in the chat we decided:
- a short-term approach could be to manaully generate svg for each chart and include, alongside a code comment containing the mermaid source
- a longer-term solution could be to create an 11ty plugin to extract the mermaid code and render it at build time, using this for guidance: https://github.com/KermanX/node-mermaid-render/

pdaoust:
Updated thoughts on this: I'm having a lot of trouble getting documents with Mermaid diagrams to go through CI. The netlify-checklinks plugin chokes on even slightly malformed SVG (which Mermaid does produce -- a non-self-closing `<br>` tag inside HTML labels, which is not valid XHTML, which is required for HTML-in-SVG) and even perfectly formed SVG in one case. The fault is shared equally by Mermaid and netlify-checklins, and here are the solutions I see:

1. Client-side mermaid (won't do unless Mermaid shrinks their bundle size)
2. Disable netlify-checklins (won't do either -- it's caught too many errors)
3. Fork netlify-checklinks and allow it to disable SVG

This problem exists even while I'm manually generating Mermaid SVG and pasting it into documents.

PR #413, opened 2024-01-08
[EPIC] Build Section

mattyg:
This is the main PR to contain many smaller PRs, before they can be merged into main.

**Overview**
- Rename "Core Concepts" to "Learn"
- Add top-level section "Build"
  - Content will focus on *practical developer usage* of holochain, with less prose, more code examples, and links directly into rust docs


pdaoust:
I like the idea of calling the new section 'build'. Fits for me. I have to confess I'm emotionally attached to the term 'core concepts', but I'm willing to change it if it's getting in the way. What's the reason for changing it to 'learn'? Consistency?

mattyg:
> I like the idea of calling the new section 'build'. Fits for me. I have to confess I'm emotionally attached to the term 'core concepts', but I'm willing to change it if it's getting in the way. What's the reason for changing it to 'learn'? Consistency?

I guess consistency and just copying language from other documenation sites, but it's not important to me -- I'd defer to you.

pdaoust:
Thanks on both counts -- your second point resolves a quandary I was just chewing on, in fact (where to put howtos like deployment, upgrade, etc)

pdaoust:
Closing because #436 deals with the CC renaming, and the Build section will automatically get created once we create our first page in that section.

pdaoust:
actually I created a skeleton in #437 without any actual individual pages.

PR #415, opened 2024-01-08
Feat/references nav and more content

mattyg:
**Overview**
- Improvements to References page:
  - standardize & modify names of sidebar links
  - add page sections navigator
  - no longer label the HDK as "HDK (Rust)" as there are currently no other HDKs available to distinguish against
  - add link to HDI and explanation text
  - add links to js & rust client docs to sidebar
  - add link to community-maintained clients (C#)
 - add link to developer immersive slides & exercises 

@pdaoust I thought about linking to existing example apps (i.e. acorn, mewsfeed), and/or to existing dev agencies (dark soil, lightningrodlabs, etc.). Do either of these make sense to you?


pdaoust:
Thanks for catching this -- I knew the section name was mismatched with the URL path, but didn't know it was also mismatched between top nav and sidebar nav. I'd suggest we do this, but rename it _all_ to 'Resources', which I feel accommodates both references and the other stuff you're adding, plus non-buildy howtos. I've just asked @MatthewNichols if that goes against some empirical findings they made when they were doing usability testing. Suspect it's okay to change though.

mattyg:
> I'd suggest we do this, but rename it all to 'Resources'

Looks like you've done this in #429 -- happy to have you merge that into this PR but lmk if you need me to.

pdaoust:
yup, I'll merge that into this once it's approved -- it's really basic but feels like it's kind of a dependency for a lot of other restructurings we want to do.

src/pages/resources/index.md
```diff
@@ -1,30 +1,48 @@
...
+    href: example-applications
+  - text: Tutorials and training
+    href: tutorials-and-training
 ---
 
-## Rust HDK
+## HDK and HDI
 
-When you write a Holochain application, the part that lives in Holochain is called a [DNA](../concepts/2_application_architecture/#layers-of-the-application-stack). It runs in a WebAssembly sandbox and talks to the host, or conductor, through the host API. The Rust HDK (Holochain Development Kit) makes it easy for you to write your DNAs in the Rust programming language.
+When you write a Holochain application, the part that lives in Holochain is called a [DNA](/concepts/2_application_architecture/#layers-of-the-application-stack). It runs as a guest in a WebAssembly sandbox and talks to the host, or Holochain conductor, through the host API. It's also expected to implement callbacks that the conductor needs to call at certain times. The HDK and HDI Rust crates make it easy for you write guest code that interfaces with the conductor --- the HDK for your DNA's [coordinator zomes](/resources/glossary/#coordinator-zome) and the HDI for [integrity zomes](/resources/glossary/#integrity-zome).
```
mattyg, line 22: I think this is TMI for this section. Once we have a Build section it can be moved there.

src/pages/resources/index.md
```diff
@@ -1,30 +1,48 @@
...
+* The **application API** lets front-ends call a running cell's functions and get information on the [DNA bundle](/resources/glossary/#dna-bundle) that the cell belongs to.
 
-For both of these APIs, you make an RPC call sending a MessagePack-serialized request to the conductor over WebSocket and listening for a response. On the interface that exposes the app API, you can also listen for [**signals**](./glossary/#signal) broadcast by cells.
+For both of these APIs, you make an RPC call sending a MessagePack-serialized request in a special envelope format to the conductor over WebSocket and listen for a response. The request's envelope must contain a request ID, and the matching response will have the same ID. On the interface that exposes the app API, you can also listen for [**signals**](/resources/glossary/#signal) broadcast by cells. There are [client libraries](#conductor-clients) for JavaScript and Rust that make it easy to handle requests/responses and set up signal listeners.
```
mattyg, line 34: I think this is TMI for this section. But lets leave it for now -- once we have a Build section it can be moved there.

src/pages/resources/index.md
```diff
@@ -34,12 +52,49 @@ The conductor has a few settings that can (and should) be configured via a YAML
...
+
+* [Holochain Open Dev](https://github.com/holochain-open-dev) has a lot of applications from the community.
+* [Holochain Foundation sample apps](https://github.com/holochain-apps) contains application written by Holochain team members.
+* [Lightningrod Labs](https://github.com/lightningrodlabs/), a group formed by Holochain co-creator [Eric Harris-Braun](https://github.com/zippy), creates foundational hApps for the Holochain ecosystem to build their tools on. [Moss](https://github.com/lightningrodlabs/we) (formerly We) is a groupware container for composing multiple applets into one cohesive experience.
+* [Glass Bead Software](https://github.com/glassbeadsoftware/) produces Snapmail, a privacy-first intranet mail app.
+* [Darksoil Studio](https://github.com/darksoil-studio/) creates apps to help groups organize and do their work together.
+* [Harris-Braun Enterprises](https://github.com/h-be) has created [Acorn](https://acorn.software/), a unique project management app based around defining goals first, then figuring out what needs to be done in order to achieve those goals.
```
mattyg, line 93: Honestly it feels a bit embarrassing that we have Eric's projects listed 3 times as different companies.

I'd rather split this into 2 lists: 
- list example apps noting the app name and what it does, and linking directly to its github. 
- another list of development agencies -- I can see that referring people to dev agencies could be valuable but also want to flag that it may become hairy at some point (i.e. why can't *my* agency be included in the list? why is this other one included? etc)

I also think mewsfeed could be a good example app to add to the list, although again a bit hairy on what gets included or now.

mattyg (commented):
Marking this as "request changes" (since I'm not actually able to do so on my own PR)

src/pages/resources/index.md
pdaoust: Take a look at the most recent commit and see what you think.

mattyg: > ha ha, I was feeling that tension myself :) With such a small ecosystem, I think we can keep this list for now and then move it to an 'awesome-holochain' list when it gets hairy (or maybe we hope for such lists from the community, and then just point to them.

yeah i think there are some from the community but not sure which is the most active or up-to-date. We should either track that down or make a new one and then link to in a subsequent PR.

> Honestly this list is already a bit patchy, and I'm worried that I'm missing apps -- but I also don't want the list to grow too much. Maybe just a few apps would be good.

Yeah makes sense to me.

> I'm also thinking that the Dev Portal isn't quite the right place to refer people to agencies; the only real reason I've included some is that they have a number of apps... and then I just continued the pattern of mentioning the agency even for individual apps.

Agreed

> Take a look at the most recent commit and see what you think.

Yeah I like this much better

mattyg (commented):
approved

PR #417, opened 2024-01-08
Feat/entries

mattyg:
Build page all about entries

**Depends on #412**


mattyg:
Responding to @pdaoust's comment from github.com/mattyg/docs-pages/pull/4

> I think we'll want to decide how long a given page should be, and how much to rely on hyperlinking instead (e.g., should updates go on the entry page? probably.

I'd say yes, unless it starts to seem like the page is too long.

> but how about the atomic commit lifecycle? maybe break that out into a separate page cuz it's going to be true for actions of all types -- that sort of thing).

Yeah makes sense.

pdaoust:
Note: I've inlined the SVG, so it no longer depends on #412 .

mattyg:
I just realized that you may not have permission to move this from draft -> PR since I'm the original author, so I just updated it to an actual pr -- not a comment on the readiness I'll leave that to you!

pdaoust:
@mattyg I can't request your review cuz this is your PR, and I've done a lot of content edits. Some things I'd like review on:

* how is the balance of prose and code? I've tried to keep the prose as terse as possible.
* Anything that looks like it won't compile? I have to admit I'm just winging the code :D
* Does the pasted mermaid SVG work for you? It ain't working for me... just a blank spot

src/pages/build/guide/index.md
```diff
@@ -0,0 +1,26 @@
...
+
+!!! note In progress
+This guide is under construction. Expect more content to be rapidly published in the first half of 2024.
+!!!
+
+::: coreconcepts-intro
+This Build Guide organizes everything you need to know about developing Holochain applications into individual topics. Each topic page stands alone as a comprehensive guide to using a given feature or implementing a given functionality. There are lots of code examples which make it clear how to do something yet are generic enough to be universally useful. These examples may not cover every single use case, though, so we'll point you to the reference documentation often.
+:::
+
+## Working with data
```
mattyg, line 13: IMO for this PR we should leave this section out. Lets open another PR to expand and clarify this content. It should also live on its own page and have a menu item under "build" -- actually probably should go first before entries.

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,397 @@
+---
```
mattyg, line 1: I think this belongs under the "build" section, not in a "guide" section.

It should be the first menu item under "build" for now.

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,397 @@
...
+    href: identifiers-on-the-dht
+  - text: Retrieve an entry
+    href: retrieve-an-entry
+  - text: Community CRUD libraries
+    href: community-crud-libraries
+  - text: Reference
+    href: reference
+---
+
+An **entry** is structured data that's serialized and written as a blob of bytes to an agent's source chain via an **entry creation action**, which can either be a `Create` or `Update` action. It can be updated or deleted. Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it. This let you distinguish separate writes of the same entry from each other.
```
mattyg, line 35:
```suggestion
An **entry** is structured data that's serialized and written as a blob of bytes to an agent's source chain via an **entry creation action**, which can either be a `Create` or `Update` action. It can be updated or deleted. Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it.
```
```suggestion
An **entry** is structured data that's serialized and written as a blob of bytes to an agent's source chain via an **entry creation action**, which can either be a `Create` or `Update` action. It can be updated or deleted. Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it. This let you distinguish separate writes of the same entry from each other.
```
```suggestion
An **entry** is structured data that's serialized and written as a blob of bytes to an agent's source chain via an **entry creation action**, which can either be a `Create` or `Update` action. It can be updated or deleted. Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it.
```

mattyg, line 35:
```suggestion
An **entry** is structured data that's serialized and written as a blob of bytes to an agent's source chain via an **entry creation action**, which can either be a `Create` or `Update` action. It can be updated or deleted. Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it.
```
I immediately found this confusing. I would just leave it out of the intro paragraph.

mattyg, line 35: > Although entry data exists as its own entity on a DHT, its associated entry creation action is semantically considered part of it and is stored along with it. This let you distinguish separate writes of the same entry from each other.

I find all of this wording confusing. I think its tmi for the intro paragraph and probably deserves to be pulled out into its own section in this file (i.e. "## Entry vs Action" or something)

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,397 @@
...
+  Movie(Movie),
+  // other types...
+}
+```
+
+Each variant in the enum holds the Rust type that defines the shape of the entry, and is implicitly marked with an `entry_def` proc macro which, if you specify it explicitly, lets you configure the given entry type further:
+
+* An entry type can be configured as **private**, which means that it's never published to the DHT, but exists only on the author's source chain. To do this, use the `visibility = "private"` argument.
+* A public entry type can be configured to expect a certain number of **required validations**, which is the number of [validation receipts](/resources/glossary/#validation-receipt) that an author tries to collect from authorities before they consider their entry published on the DHT. To do this, use the `required_validations = <num>` argument.
+* You can override the name of an entry type, which defaults to the name of the enum variant.
```
mattyg, line 74: I think we should leave the name overriding out -- i honestly don't see any utility for it versus just creating another enum type.

mattyg:
>     * how is the balance of prose and code? I've tried to keep the prose as terse as possible.

For me personally its a *lot* of prose. If I'm reading docs I would rather it be oversimplified prose with code examples, and push the deeper explanations into "Advanced" sections. 

That said you've added a incredible amount of valuable info. I'd way rather just get this deployed and revisit that later.

>     * Does the pasted mermaid SVG work for you? It ain't working for me... just a blank spot

It does not for me either. If you go to https://mermaid.live, disable "Pan & Zoom", then click Actions > SVG, it generates a working svg. Here's an example: https://mermaid.live/edit#pako:eNptz8tuwjAQBdBfse46RONH7OBFpQL9grabYhYWcQuCPOQmUmmUf6-JlFXZzcwZXc2MOLZVgMVX9N2JvW1cw9hm_95Vvg-MH9hq9cSe99sYUn-443ZB8QB3C8oH-LKg-ofIUIdY-3OVThnvyw79KdTBwaay8vHi4Jop7fmhb19vzRG2j0PIMMyZu7NPH9Swn_76naadb2BH_MAKWueapORGaDJKiAw3WG6KvBCGqKRCa0VaTBl-2zYlUG5kyUlxaUirtSzVHPcx4xw__QHCaVxb

Also I think we should either include the mermaid code in an html comment, or link to the mermaid.live you used to produce (probably the latter is simpler), so we can modify later.

mattyg (commented):
requesting changes about mermaid rendering and build / guide setup, but generally thrilled to see this and get it deployed! I'll try to do another read through later.

src/pages/build/guide/index.md
pdaoust: Weird, I don't know why GH is seeing that as an addition -- it already exists on the build guide page.

Re: menu sub-item, thanks for the reminder. I need to make that happen too, maybe in another PR.

src/pages/build/guide/entries.md
pdaoust: I've got plans for other sub-sections. Does it seem weird to have this as the only sub-section?

src/pages/build/guide/entries.md
pdaoust: good point. I was just trying to be thorough ;)

pdaoust:
Thanks so much for reviewing this. I'll work on some of the content, and I'm also happy to remove some of the stuff. I'm thinking about taking out the part about updates too -- super useful info, but can go into the eventual 'Build > Patterns' section (or 'Build > Cookbook' or whatever we wanna call it)

pdaoust:
(BTW thanks for the tip about Mermaid -- the navigation stuff is indeed what's causing the problem. Strangely, I think it was `mmdc` that generated that, and I expected it to only generate plain SVG)

mrruby (approved):

src/pages/build/guide/entries.md
mattyg: Yes it's weird to have this as the only section, and makes it too hard to discover the entries page. 

Generally if we want to have sections nested 2 levels deep we need the sidebar menu to support displaying 2 levels of depth.

(i.e. entries should appear here):
image

src/pages/build/guide/index.md
mattyg: No this "working with data" section is an addition in this PR. This is what the build guide index page looks like currently:

image

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+5. Wait for the zome function to complete.
+6. Convert the action to DHT operations.
+7. Run the validation callback for all DHT operations.
+    * If successful, continue.
+    * If unsuccessful, return the validation error to the client instead of the zome function's return value.
+8. Publish the actions in the scratch space to the source chain.
+9. Return the zome function's return value to the client.
+10. In the background, publish all newly created DHT operations to their respective authority agents.
+
+<!-- TODO review and outline steps that are taken under the hood *exactly*, including which DHT ops are published -->
```
mattyg, line 157:
```suggestion
```

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
```
mattyg, line 204:
```suggestion
```

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Update patterns
+
+Holochain gives you this `update_entry` function, but is somewhat unopinionated about how it's used. You can interpret an update as applying to either the original _action_ or the original _entry_ being updated, because the `Update` action is merely a piece of metadata attached to both, and can be retrieved along with the original data.
+
+You can also choose where to attach updates. You can structure them as a 'list', where all updates refer to the `ActionHash` of the original `Create` action.
+
+<!-- TODO: Replace the SVG with Mermaid once we've got build-time Mermaid rendering working.
+I can't keep the original Mermaid in a comment because the arrows in the code break the comment tag early.
```
mattyg, line 213: Lets include a mermaid.live URL instead. It doesn't break out of the html comment:

https://mermaid.live/edit#pako:eNptz8tuwjAQBdBfse46RONH7OBFpQL9grabYhYWcQuCPOQmUmmUf6-JlFXZzcwZXc2MOLZVgMVX9N2JvW1cw9hm_95Vvg-MH9hq9cSe99sYUn-443ZB8QB3C8oH-LKg-ofIUIdY-3OVThnvyw79KdTBwaay8vHi4Jop7fmhb19vzRG2j0PIMMyZu7NPH9Swn_76naadb2BH_MAKWueapORGaDJKiAw3WG6KvBCGqKRCa0VaTBl-2zYlUG5kyUlxaUirtSzVHPcx4xw__QHCaVxb

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+
+Or you can structure your updates as a 'chain', where each update refers to the `ActionHash` of the previous entry creation action (either an `Update` or the original `Create`).
+
+<!-- TODO: This one too. The original graph was TB (top to bottom), but RL (right to left) looks better in the space. -->
```
mattyg, line 220: lets include a mermaid.live url in the comment -- we'll appreciate it later

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+4. Wait for the zome function to complete.
+5. Convert the action to DHT operations.
+6. Run the validation callback for all DHT operations.
+    * If successful, continue.
+    * If unsuccessful, return the validation error to the client instead of the zome function's return value.
+7. Publish the actions in the scratch space to the source chain.
+8. Return the zome function's return value to the client.
+9. In the background, publish all newly created DHT operations to their respective authority agents.
+
+<!-- TODO review and outline steps that are taken under the hood *exactly*, including which DHT ops are published -->
```
mattyg, line 272:
```suggestion
```

mjbrisebois (approved):
LGTM.  Just the 1 suggestion.

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+
+If you structure your updates as a chain, you may want to also create links from the `ActionHash` of the original `Create` to each update in the chain, making it a hybrid of a list and a chain. This trades additional storage space for reduced lookup time.
+
+### Resolving update conflicts
+
+It's up to you to decide whether two updates on the same entry or action are a conflict. If your use case allows branching edits similar to Git, then conflicts aren't an issue.
+
+But if your use case needs a single canonical version of a resource, you'll need to decide on a conflict resolution strategy to use at retrieval time.
+
+If only the original author is permitted to update the entry, choosing the latest update is simple. Just choose the `Update` action with the most recent timestamp, which is guaranteed to [advance monotonically](https://doc.rust-lang.org/std/time/struct.Instant.html) for any honest agent. But if multiple agents are permitted to update an entry, it gets more complicated. Two agents could make an update at exactly the same time (or their action timestamps might be wrong or falsified). So, how do you decide which is the 'latest' update?
```
mattyg, line 231:
```suggestion
If only the original author is permitted to update the entry, choosing the latest update is simple. Just choose the `Update` action with the most recent timestamp, which is guaranteed to advance monotonically for any honest agent. But if multiple agents are permitted to update an entry, it gets more complicated. Two agents could make an update at exactly the same time (or their action timestamps might be wrong or falsified). So, how do you decide which is the 'latest' update?
```
don't think it adds anything to link to this

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+
+These are two common patterns:
+
+* Use an opinionated, deterministic definition of 'latest' that can be calculated from the content of the update.
+* Model your updates with a data structure that can automatically merge simultaneous updates, such as a [conflict-free replicated data type (CRDT)](https://crdt.tech/).
```
mattyg, line 236: maybe we should add crdt.tech to the resources page? Or link to it near the references section of this page? It's a great resource and maybe should be more clearly noted than an inline link (which I assume just links to a definition when reading it).

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+    * If unsuccessful, return the validation error to the client instead of the zome function's return value.
+8. Publish the actions in the scratch space to the source chain.
+9. Return the zome function's return value to the client.
+10. In the background, publish all newly created DHT operations to their respective authority agents.
+
+<!-- TODO review and outline steps that are taken under the hood *exactly*, including which DHT ops are published -->
+
+### Update patterns
+
+Holochain gives you this `update_entry` function, but is somewhat unopinionated about how it's used. You can interpret an update as applying to either the original _action_ or the original _entry_ being updated, because the `Update` action is merely a piece of metadata attached to both, and can be retrieved along with the original data.
```
mattyg, line 208: > You can interpret an update as applying to either the original _action_ or the original _entry_ being updated, because the `Update` action is merely a piece of metadata attached to both, and can be retrieved along with the original data.

This just adds confusion for me. I just want to be told how I *should* think about it. As far as I understand an Update action is always applied to another Update or Create action. It is never applied to an entry.


I might just leave this out.

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+* A common pattern to identify an _instance_ of an entry (i.e., an entry authored by a specific agent at a specific time) is to use the `ActionHash` of its entry creation action instead. This gives you timestamp and authorship information for free, and can be a persistent way to identify the initial entry at the root of a tree of updates.
+* You can reference an agent via their `AgentPubKey`. This is a special type of DHT entry whose identifier is identical to its content --- that is, the agent's public key. You can use it just like an `EntryHash` and `ActionHash`.
+* Finally, you can also use **external identifiers** (that is, IDs of data that's not in the DHT) as long as they're 32 bytes, which is useful for hashes and public keys. It's up to you to determine how to handle these identifiers in your front end.
+
+You can use any of these identifiers as a field in your entry types to model a many-to-one relationship, or you can use links between identifiers to model a one-to-many relationship.
+
+## Retrieve an entry
```
mattyg, line 285: Thinking maybe we should make sub headings for "Only Entry" and "Entry Details" -- I think it makes it clearer to split this section and to have both links in the sidebar

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+            None => debug!("Movie entry couldn't be retrieved"),
+        }
+    }
+    _ => debug!("Movie entry not found"),
+}
+```
+
+## Community CRUD libraries
+
+If the scaffolder doesn't support your desired functionality, or is too low-level, there are some community-maintained libraries that offer opinionated and high-level ways to work with entries. Some of them also offer permissions management.
```
mattyg, line 395: This is the first time the scaffolder is mentioned in this page. Maybe this section should be renamed "CRUD Scaffolding" and note that the scaffolder will generate this code for you, or alternatively there are some community libraries.

(I realize I wrote this but good to re-read with fresh eyes).

src/pages/build/guide/entries.md
```diff
@@ -0,0 +1,408 @@
...
+- [hc-cooperative-content](https://github.com/mjbrisebois/hc-cooperative-content)
+
+## Reference
+
+- [hdi::prelude::hdk_entry_helper](https://docs.rs/hdi/latest/hdi/attr.hdk_entry_helper.html)
+- [hdi::prelude::hdk_entry_defs](https://docs.rs/hdi/latest/hdi/prelude/attr.hdk_entry_defs.html)
+- [hdi::prelude::entry_def](https://docs.rs/hdi/latest/hdi/prelude/entry_def/index.html)
+- [hdk::prelude::create_entry](https://docs.rs/hdk/latest/hdk/entry/fn.create_entry.html)
+- [hdk::prelude::update_entry](https://docs.rs/hdk/latest/hdk/entry/fn.update_entry.html)
+- [hdi::prelude::delete_entry](https://docs.rs/hdk/latest/hdk/entry/fn.delete_entry.html)
```
mattyg, line 408: just a warning that linking to /latest/ on the rust docs is going to bite us. We'll want to prioritize versioning soon #432

mattyg (commented):
alright i added a bunch more comments. After those I'm completely satisfied, feel free to merge when you are. Feeling really really good about this. Thanks for all your work!

src/pages/build/guide/entries.md
mattyg: Actually re-reading this I think it's important information to keep, but I do think its a bit confusing. Also maybe it fits better in the previous section "updates under-the-hood" rather than the intro to "update patterns"?

> interpret an update as applying to either the original _action_ or the original _entry_ being updated

This seems misleading. My mental model is that updates always apply to actions. That's structure of them on the source chain as well.

> the `Update` action is merely a piece of metadata attached to both,

This seems misleading. The update action is attached to the updated entry, but *not* to the original entry -- instead its attached to the original create entry *action*. 

>  and can be retrieved along with the original data.

When you get an entry you receive the oldest-live record, maybe not the one you care about.

PR #421, opened 2024-02-02
instructions on getting a specific holonix version

pdaoust:
(plus a capitalisation fix)

### DIFF
--- a/src/pages/get-started/install-advanced.md
+++ b/src/pages/get-started/install-advanced.md
@@ -25,19 +25,19 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
 nix develop github:holochain/holochain#holonix
 ```
 
-To get ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
+To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
 
 ```shell
-nix develop --override-input versions ./versions/0_2 .#holonix
+nix develop --override-input versions "github:holochain/holochain?dir=versions/0_2" "github:holochain/holochain#holonix"
 ```
 
 And this example gives you the newest weekly developer snapshot:
 
 ```shell
-nix develop --override-input versions ./versions/weekly .#holonix
+nix develop --override-input versions "github:holochain/holochain?dir=versions/weekly" "github:holochain/holochain#holonix"
 ```
 
-Take a look at the [`versions/` folder in the `holochain/holochain` repository](https://github.com/holochain/holochain/tree/develop/versions) to find out what versions you can target. Each subfolder is a valid version number to use in the above command, and the `flake.nix` file inside the subfolder shows the specific release tag for each component that will be downloaded.
+Take a look at the [`versions/` folder in the `holochain/holochain` repository](https://github.com/holochain/holochain/tree/develop/versions) to find out what versions you can target. Each subfolder is a valid version to use in the above command, and the `flake.nix` file inside the subfolder shows the specific release tags that will be used for Holochain and the Lair Keystore. The launcher and scaffolding tools live in separate repos, so you will get the latest release of those tools that are available on their respective release branches referenced in the version-specific `flake.nix` file. Note that you may temporarily see older versions than the current tip of the referenced release branch because our automation updates the `flake.lock` separately, so that file is the absolute source of truth for what versions you will get.
 
 #### Enabling Flake features on your system
 
--- a/src/pages/get-started/install-advanced.md
+++ b/src/pages/get-started/install-advanced.md
@@ -25,13 +25,13 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
 nix develop github:holochain/holochain#holonix
 ```
 
-To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
+By default, the above one-liner will give you the latest recommended release (currently from the 0.2.x series). To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the next release candidate in the 0.2.x series:
 
 ```shell
-nix develop --override-input versions "github:holochain/holochain?dir=versions/0_2" "github:holochain/holochain#holonix"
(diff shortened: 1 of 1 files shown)


steveej (approved):
:raised_hands:

steveej (dismissed):
> 🙌

ops, i was too quick. i realize now that the context of these commands are _not_ in a checkout of the holochain repository. hence, the version paths which assume that won't work in any other case. solutions are either make that context explicit or give absolute URLs to github.

src/pages/get-started/install-advanced.md
```diff
@@ -25,6 +25,20 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
...
+nix develop --override-input versions ./versions/0_2 .#holonix
+```
+
+And this example gives you the newest weekly developer snapshot:
+
+```shell
+nix develop --override-input versions ./versions/weekly .#holonix
+```
+
+Take a look at the [`versions/` folder in the `holochain/holochain` repository](https://github.com/holochain/holochain/tree/develop/versions) to find out what versions you can target. Each subfolder is a valid version number to use in the above command, and the `flake.nix` file inside the subfolder shows the specific release tag for each component that will be downloaded.
```
ThetaSinner, line 40:
```suggestion
Take a look at the [`versions/` folder in the `holochain/holochain` repository](https://github.com/holochain/holochain/tree/develop/versions) to find out what versions you can target. Each sub-directory is a valid version to use in the above command, and the `flake.nix` file inside the sub-directory shows the specific release tags that will be used for Holochain and the Lair Keystore. For the launcher and scaffolding, you will get the latest release of those tools that are available on the release branch referenced in the `flake.nix`. Note that you may temporarily see older versions than the `flake.nix` suggests because our automation updates the `flake.lock` separately, so that file is the absolute source of truth for what versions you will get.
```

ThetaSinner: This is wordy but contains more details. Up to you what you want to do with it! Maybe you can shorten it a bit while keeping the important details?

pdaoust:
Darn, I was hoping nix was more magical than that. (thanks @ThetaSinner for catching the complete absence  of `github:holochain/holochain` in my commands; don't know how that slipped past me. Must've copy/pasted the nearest command w/o paying attention.)

src/pages/get-started/install-advanced.md
pdaoust: Wordy is good. This is a page for devs who are looking for details, so I think we should keep it all in. I'm a bit torn on subfolder vs subdirectory though; I've used 'folder' everywhere else in the documentation. Thoughts on whether 'directory' is an important distinction to make?

ThetaSinner: Not important, happy to drop that from my suggestion

ThetaSinner: Done

ThetaSinner:
It's open source, any magic you feel is missing can be added. You could create a version that always assumes relative paths point to Holochain :) But yeah, it's simpler to reference derivations from inside the repo than from outside unfortunately. Nothing too crazy syntax-wise at least, just some long commands.

Something else it would be good to document somewhere at some point is an answer to the questions "You modified Holochain for me, how do I run that from your branch?" and "You modified the HDK for me, how do I use that in my happs before you merge your changes". These are questions I've had several times, and the answer is often "The change is going into develop so you need a happ based on 0.3.x to be able to use it" but it does happen that people running on 0.3.x would be happy to test things, but it's not obvious how to do it. So when I say we should document it, what I actually mean is that somebody should push figuring out how to the top of my priority list so we can document it 😁

src/pages/get-started/install-advanced.md
pdaoust: @ThetaSinner I've edited it a bit for clarity; did I get the details right? As I understand this, launcher and scaffolding have branches that maintain compatibility with a Holochain major release series, but the `flake.lock` might cause Nix to pull in a slightly older commit on that branch if CI hasn't refreshed `flake.lock` lately?

src/pages/get-started/install-advanced.md
```diff
@@ -25,6 +25,20 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
 nix develop github:holochain/holochain#holonix
 ```
 
+To get ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
```
pdaoust, line 28:
```suggestion
To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
```

src/pages/get-started/install-advanced.md
ThetaSinner: Yes I think that is all accurate and clear but because of the ordering it might sound like only scaffolding and launcher can be behind. The same is also true of Holochain and Lair Keystore. If I update the `flake.nix` to use the `0.2.6` tag for Holochain but the `flake.lock` will still be pointing at `0.2.5` until the automation bumps it. It runs every 6 hours currently so it's not a long lag but if the CI fails then it can run a few days behind until we get it fixed.

ThetaSinner:
@pdaoust Do you have what you need from me to finish this off? I'd love to see this deployed so I can point people at it :)

pdaoust:
@ThetaSinner I do -- I'm going to change the examples to reflect that 0.2 is the default now though.

src/pages/get-started/install-advanced.md
```diff
@@ -25,6 +25,20 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
 nix develop github:holochain/holochain#holonix
 ```
 
+To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`. This example gives you the newest release in the unstable 0.2.x series:
+
+```shell
+nix develop --override-input versions "github:holochain/holochain?dir=versions/0_2" "github:holochain/holochain#holonix"
```
pdaoust, line 31:
```suggestion
nix develop --override-input versions "github:holochain/holochain?dir=versions/0_1" "github:holochain/holochain#holonix"
```

src/pages/get-started/install-advanced.md
```diff
@@ -25,6 +25,20 @@ The flake-based one-liner to get you an ad-hoc Holonix shell looks like this:
...
+```
+
+And this example gives you the newest weekly developer snapshot:
```
pdaoust, line 34:
```suggestion
And this example gives you the newest weekly developer snapshot (currently in the 0.3.x series):
```

ThetaSinner (approved):

PR #423, opened 2024-02-05
add cspell GH workflow for spell checking

pdaoust:
Came from here: https://github.com/marketplace/actions/cspell-action


mattyg (approved):
great!

pdaoust:
Thanks for reviewing @mattyg ; merging now.

PR #428, opened 2024-02-23
Removing Design System from Site Search and updated Pagefind lib

MatthewNichols:


pdaoust (approved):
I don't think I know enough about Pagefind and the site structure to know 100% whether this PR is okay, so I'm gonna trust your wisdom on that @MatthewNichols ! :D I'm guessing the design system files don't use `src/pages/_includes/base-layout.njk` and are therefore not indexed?

MatthewNichols:
> I don't think I know enough about Pagefind and the site structure to know 100% whether this PR is okay, so I'm gonna trust your wisdom on that @MatthewNichols ! :D I'm guessing the design system files don't use `src/pages/_includes/base-layout.njk` and are therefore not indexed?

That is correct. I added a section to the ReadMe that goes into it a little, and points to the documentation for more in-depth.

PR #429, opened 2024-02-29
change /references path to /resources, along with all labels

pdaoust:


pdaoust:
Skipping review from Mary, cuz it's basically just implementing a change she said ought to have been made consistently throughout the site. My only concern is losing SEO/bookmarks/inbound links, which I assume will be covered by the 301 redirects I set up.

PR #431, opened 2024-03-04
build: use rollup to build typescript, minify js bundles

mattyg:
- use rollup to build typescript
- use rollup to minify js bundles with terser


MatthewNichols (approved):
Looks good. I definitely learned from this. Thank you.

PR #433, opened 2024-03-12
Rename "Core Concepts" to "Learn"

pdaoust:
making this into its own PR based on @mattyg 's suggestion in #413 . Rationale: 'Learn' is more consistent with 'Get Started' and the future 'Build', and it may be more consistent with how other dev docs are structured too.


src/custom/style.css
```diff
@@ -149,21 +149,21 @@ article img { display: table; margin-left: auto; margin-right: auto; }
 
 /* Core Concepts */
```
mattyg, line 150:
```suggestion
/* Learn */
```

mattyg (approved):

pdaoust:
I'm having second thoughts about this -- instead, I think I'll just rename it in the menus and URLs, but keep the name 'Core Concepts' in the page content. I have plans for other stuff -- video content (which I don't want to do but I know people want), maybe more tutorials which are sorta buildy but more properly learny. Therefore, 'Core Concepts' seems useful as a header for the abstract stuff.

pdaoust:
Picking this up in #436 -- in which I only change the menu, not even the URLs.

PR #434, opened 2024-03-15
WIP - first try at integrating Mermaid into CI

pdaoust:
Just working away at it... this integrates `mermaid-cli` right into 11ty and markdown-it, so we'll see if it works!


pdaoust:
Putting this on pause for now -- there's a lot of async stuff going on in `mermaid-cli`'s Node.JS API, and that doesn't play nice with markdown-it's CommonJS-ish-ness.

pdaoust:
not working -- too many async-to-sync function conversions that I just can't figure out. Mermaid is all async; markdown-it is all sync.

PR #435, opened 2024-03-18
Spike: update 11ty to 3.x

pdaoust:
* Convert all the site's JS files to ES6 modules
* Write a helper function to import JSON5 (no longer using `require`, and registering a `require`-able extension via `module.extension` is deprecated anyway)
* Rename `.eleventy.js` to `eleventy.config.js`
* Rename rollup config file


pdaoust:
CI is red, but I updated the Netlify config to use Node.js 18, retried the deploy, and it succeeded. Guess retries don't show up in GH checks.

mattyg (approved):

PR #438, opened 2024-03-25
Remove 'coreconcepts-' prefix from styles, admonition blocks, etc

pdaoust:
There are some template elements called `coreconcepts-intro`, `coreconcepts-orientation`, etc. The naming is a legacy from when they were only used on the core concepts pages, but now they're getting used in other places too.

There are a couple places where they show up:

* The Markdown parser has been extended to recognise them as special blocks that deserve their own styling
* There are matching CSS styles

~~This is just a simple search-and-replace.~~ _Update: see below_. I'm pretty sure the shortened names don't clash with anything.

## Update

I found out the styles weren't hooked up to these block containers anyway, so I've done some work to make them work. You can see all three at work in e.g., https://deploy-preview-438--developer-portal-production.netlify.app/concepts/7_validation/

**One big change** is that the big `::: storysequence` block has been changed to more verbose, but much simpler and less magical, `::: storystep` blocks around each step in a story sequence. I don't think we'd ever reimplemented the original site's behavour, in which each step was a list item.


pdaoust:
okay, this is weird... I just discovered that these special Markdown blocks get custom markup and styles, but never the twain do meet -- the styles live in a file that never gets loaded by any page!

src/scss/_blocks.scss
```diff
@@ -0,0 +1,28 @@
+.intro {
+  font-size: $fs-even-bigger;
```
mattyg, line 2: font feels a bit too large to me

image

src/scss/_blocks.scss
```diff
@@ -0,0 +1,28 @@
...
+  border-left: 4px solid $cl-gray;
+  padding-left: 1em;
+  font-weight: 500;
+  line-height: 1.55;
+}
+
+.orientation, .storystep {
```
mattyg, line 9: I like the added background color to distinguish the content. I think the text could be distinguished a bit more clearly too? Something like this:

image

mattyg (approved):
Noted a few potential styling tweaks (I don't have a good design sense so feel free to ignore).

Also I see that the Get Started page and Resources page have "intro" text that isn't tagged for the intro styling.

src/scss/_blocks.scss
pdaoust: I might eventually apply this to all images, actually -- we're working with a graphic designer who's added a gray background to all her example illustrations and it looks spiffy. I feel like the box around the text may add too much differentiation... maybe just a little horizontal line. Tough to do this in CSS, but I might be able to.

src/scss/_blocks.scss
pdaoust: fair feedback... guess the large font won't look good for every intro block. I'm personally okay with this look, but if it looks jarring to you I can play with sizes.

src/scss/_blocks.scss
mattyg: sure -- you can do "border-top" on the text element for a horizontal line

src/scss/_blocks.scss
mattyg: eh nah lets leave it for a real designer down the road.

pdaoust: I used to be one :)

src/scss/_blocks.scss
pdaoust: check out core concept seven -- I played around with `<hr/>` elements, so the author can opt in if they like the look and opt out if they don't. Could lead to inconsistency, but I'd rather inconsistency that fit the need than consistency that looked awkward in places.

https://deploy-preview-438--developer-portal-production.netlify.app/concepts/7_validation/

src/scss/_blocks.scss
mattyg: ha! well than I defer to you

src/scss/_blocks.scss
pdaoust: updated: I've applied the treatment to all story steps. (yay regexes)

mattyg: nice - I prefer it with the original margin, i.e.
image

but also I'm not a designer and defer to you

PR #440, opened 2024-03-26
add icon to external links

pdaoust:
Add a little icon to any link that points offsite. It can be overridden by adding a `.no-external-link-icon` class to a link or its container, and a container-wide override can be overridden by adding a `.external-link-icon` class to a child link.

You can take a look at pretty much any page in the deploy preview to see what this looks like; e.g, the Getting Started page https://deploy-preview-440--developer-portal-production.netlify.app/get-started/


mrruby (approved):

mattyg (changes requested):
Good idea

Looks like its not being applied to a bunch of the links on the Resources page?
image

Also the styling of the external link image looks great within the content, but slightly off for external links in the sidebar (either enlarge or vertically center):
image

pdaoust:
@mattyg this is why we do peer review; thanks! I didn't even consider the consequence of this change re: menus.

pdaoust:
huh, that issue you pointed out @mattyg is present on the existing site -- which caused me to discover the existing styling for the same thing I was trying to do in main content, which allowed me to DRY the code up a fair bit.

pdaoust:
@mattyg could you take another look and see if I've fixed the issues you brought up?

PR #441, opened 2024-03-28
Header anchors

pdaoust:
We're using the [`markdown-it-anchor`](https://www.npmjs.com/package/markdown-it-anchor) plugin to get `id` attributes on all the headers. This PR wraps the header text in an `<a>` and adds styling to make a header not obviously a link until you hover over it. You can see something similar on https://developer.mozilla.org .


mattyg (approved):
great!

PR #448, opened 2024-04-15
Feat/auto-generated TOCs

pdaoust:
This feature adds an auto-generated table of contents to every Markdown page, unless disabled by the page's front matter with `tocData: false` or an explicit TOC value for `tocData`. Comes with documentation and everything.


pdaoust:
Merging without sufficient oversight because I don't know who I'd get to review this PR :D

PR #450, opened 2024-05-16
hotfix: make it clearer that the Launcher can host both hApps and web hApps

pdaoust:
... and explain how to install the latter.


pdaoust:
gonna merge without review; hope nobody minds :)

PR #452, opened 2024-05-22
add instructions for fixing security error in Ubuntu 24.04

pdaoust:


src/pages/get-started/2-hello-world.md
```diff
@@ -30,6 +30,17 @@ nix develop
 ```shell
 npm install
 ```
+
+!!! info "Warning for Ubuntu 24.04 and later
+Ubuntu Linux 24.04 [introduces security policy changes](https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890#unprivileged-user-namespace-restrictions-15) that cause the next command to fail. Here's a simple fix. In your terminal, run this command:
```
matthme, line 35: Would a small note be useful that this is a one-time thing and doesn't need to be re-run every time before running `npm start`? Might be obvious though...

matthme (approved):
Small optional suggestion but otherwise looking great.

src/pages/get-started/2-hello-world.md
pdaoust: Good idea @matthme . What's obvious to us isn't obvious to everyone; this'd be a courteous change to make.

PR #457, opened 2024-06-24
add note about installing ubuntu packages for WSL2

pdaoust:


src/pages/get-started/index.md
```diff
@@ -47,6 +47,12 @@ For Windows users, please note that the Nix package manager, which is used to in
 
 Holochain is supported in WSL2 via the Ubuntu distribution.
 
+You'll also need to install a few packages if you want to run two dev tools, `hc launch` and `hc spin`, which start your app's back end and open its GUI in Tauri or Electron webviews:
```
ThetaSinner, line 50: These are only needed for `hc-launch`. I never had problems with Tauri needing libraries.

`adwaita-icon-theme` solves a Tauri specific problem that Electron doesn't seem to have.

I get grouping them together to get people set up quickly. I wonder if it's worth keeping a note somewhere for ourselves about why we're recommending which libraries? I don't know if we can comment in the markdown here without it showing up in the generated page?

ThetaSinner (approved):

ThetaSinner:
Nice! Thanks for adding this :)

src/pages/get-started/index.md
pdaoust: Sure, and I can clarify in the advanced install guide.

src/pages/get-started/install-advanced.md
```diff
@@ -290,3 +290,19 @@ There are other fixes [outlined in the Ubuntu 24.04 release notes](https://disco
...
+
+There are two dev tools, `hc launch` and `hc spin`, which start your app's back end and open its GUI in [Tauri](https://tauri.app/) or [Electron](https://www.electronjs.org/) webview containers, respectively. Because the Ubuntu OS installed from the Microsoft Store doesn't come with GUI packages by default, you'll need to install just a few in order to get these tools to work.
+
+If you're only using `hc launch`, which uses the Tauri webview, install this package, which resolves a GDK error about cursors:
+
+```shell
+sudo apt install -y adwaita-icon-theme
+```
+
+If you're only using the more modern `hd spin`, which uses the Electron webview, install these missing packages that are needed by the `chrome-sandbox` binary:
```
pdaoust, line 304:
```suggestion
If you're only using the more modern `hc spin`, which uses the Electron webview, install these missing packages that are needed by the `chrome-sandbox` binary:
```

PR #459, opened 2024-07-01
Recommend cargo update in upgrade guides

ThetaSinner:
I've had two reports now for people upgrading versions that the `time` package being on an older version has caused their WASM to not build. The solution to this has just been to `cargo update`, so I thought I'd add a section to the upgrade guides to mention this.


pdaoust (approved):
Approved with some formatting fixes of my own in a follow-up commit.

PR #486, opened 2024-10-25
WIP: build validation page

### LINKED ISSUE #468: Practical docs for using Validation
Closed by #520.

Must include:
- code examples
- practical, terse explanation

Topics to include (consider splitting into sub-pages):
* Validation
    * Determinism
        * dependencies
            * `must_get_*` functions
                * `must_get_valid_record` is the only one that checks for validation receipts/warrants because an entry is only valid/invalid in the context of an action
                * this also means it trusts the word of others (1-of-n validation)
            * short-circuiting with `?`
    * Operate on ops
        * flat ops convenience thingy
    * Considerations re: what should be validated on each op (IOW, what authorities should be responsible for what things)
    * Things that can be validated
        * membrane proof
        * entry structure (`entry_def` macro gives you deserialisation and error short-circuiting for free with `?` operator)
        * permission
        * rate limiting with `weight` field
        * dependencies, incl source chain history
            * Inductive validation for costly dep trees (pattern?)
    * Limitations
        * Cannot `must_get` links or actions on a base
        * Cannot currently co-validate multiple actions (can only validate an action based on prior valid actions)
        * Cannot validate the non-existence of something, because that can always change
    * (future) source chain restructured to atomic bundle of actions, co-validated
    * sys validation
    * Lifecycle of a validation
        * At author time
        * At publish time
        * At gossip time
* Sys validation
    * What does it validate?
        * Source chain continuity
        * Timestamp monotonicity
        * Contiguity of authorship
            * (future) if DPKI enabled, sys validation checks that changes of agent ID are reflected in DPKI DNA as-at action timestamp
        * What else?

mattyg:
Resolves #468

### DIFF
--- a/src/pages/build/index.md
+++ b/src/pages/build/index.md
@@ -17,4 +17,5 @@ This Build Guide organizes everything you need to know about developing Holochai
 
 * [Overview](/build/working-with-data/) --- general concepts related to working with data in Holochain
 * [Entries](/build/entries/) --- creating, reading, updating, and deleting
+* [Validation](/build/validation/) --- validating changes to DHT state
 :::
--- a/src/pages/build/validation.md
+++ b/src/pages/build/validation.md
@@ -1,0 +1,167 @@
+---
+title: "Validation"
+---
+
+::: intro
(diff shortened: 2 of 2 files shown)


pdaoust:
Closing cuz I used it as a starting point for #520.

PR #499, opened 2024-11-22
Holochain 0.4 upgrade guide

pdaoust:
Incorporates all the breaking changes in 0.4 and peripheral tooling.


src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,111 @@
+---
```
pdaoust, line 1: The URLs in this document need to be updated once the Rustdoc is published to docs.rs

pdaoust (commented):
don't deploy until 0.4 is live and recommended.

src/pages/get-started/install-advanced.md
```diff
@@ -19,18 +19,26 @@ If you want to learn more about how this setup works and how to create it manual
...
 ```shell
 nix develop github:holochain/holonix
 ```
 
+#### Specifying a certain release
+
 The above one-liner will give you the latest version of Holochain from branch `main`. To get an ad-hoc shell with a specific version of Holochain, use the flag `--override-input versions <version_path>`.
 
 ```shell
 nix develop --override-input holochain "github:holochain/holochain?ref=main-0.4" github:holochain/holonix
```
matthme, line 35: I just tried this command like so and it seemed to work as well:
```
nix develop github:/holochain/holonix?ref=main-0.4
```
If someone can verify this I think this would be much cleaner.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,111 @@
...
+:::
+
+## Summary
+
+Here are all the breaking changes you need to know about in order to update your app for Holochain 0.4:
+
+* Some unstable features are now behind feature flags.
+* The `OpenChain` and `CloseChain` actions have been modified.
+* The database encryption scheme has changed, and the directory structure has moved around.
+* The `InstallApp` admin API endpoint has had its request payload changed.
```
matthme, line 18: I wonder whether it would be worth mentioning here that the dna properties can now be provided at install time [see changelog](https://github.com/holochain/holochain/blob/main-0.4/CHANGELOG.md#20241125215336). Or mention that further down where it's currently only mentioned that the agent key is now optional in the install app payload.

src/pages/get-started/install-advanced.md
ThetaSinner: What's here is actually wrong and is likely to break. That's asking for the latest version of Holonix, which could be supporting any Holochain version and then injecting some version of Holochain into it, without matching the Lair version or anything like that.

I'm assuming we hadn't created the `main-0.4` branch when this was written. 

It is valid to do version overrides, just not like this. For example, the following one-liner would downgrade you to Holochain `0.4.0-rc.1`

```suggestion
nix develop --override-input holochain "github:holochain/holochain?ref=holochain-0.4.0-rc.1"  "github:/holochain/holonix?ref=main-0.4"
```

However, it gives you `lair-keystore` at 0.5.3 where Holochain `0.4.0-rc.1` required `lair-keystore` at 0.5.2... So this is really opening a can of worms. Same is true for `hc-launch` and `hc-scaffold`, yes they will be provided to you, but if you mismatch the versions that Holonix is giving you then compatibility becomes your problem.

src/pages/get-started/install-advanced.md
```diff
@@ -19,18 +19,26 @@ If you want to learn more about how this setup works and how to create it manual
...
 ```
 
+### Customizing the Holochain binary
+
+If you want to enable or disable certain Holochain features, such as unstable features, it's best to do this in a local flake file. [Read the 'Customized Holochain build'](https://github.com/holochain/holonix?tab=readme-ov-file#customized-holochain-build) on the Holonix readme to find out how. Keep in mind that, because you'll be creating a custom Holochain binary, you won't be able to take advantage of the package cache, so it'll take a while to compile Holochain on your machine.
```
ThetaSinner, line 40: That's only part of the problem. It also means that if you're using the GitHub Actions workflow provided by scaffolding which does `nix develop` then your CI will be slow because it will build Holochain every time.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,111 @@
...
+
+In your integrity zome's validation functions, you deal with DHT operations, or ops. They are somewhat complex, so [`FlatOp`](https://docs.rs/hdi/latest/hdi/flat_op/enum.FlatOp.html) was introduced to make things simpler. It was originally called `OpType`, and until now that old name was a deprecated alias of `FlatOp`. The old type has finally been removed, along with the `Op::to_type` method (use [`OpHelper::flattened`](https://docs.rs/hdi/latest/hdi/op/trait.OpHelper.html#tymethod.flattened) instead).
+
+## Change in enum serialization
+
+The default serialization for bare enum variants has changed. Previously, they would look like this (JSON representation):
+
+```json
+{
+    "variant1": null
```
mattyg, line 101:
```suggestion
    "MyEnumType": {"Variant1": null}
```

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+Now they look like this:
+
+```json
+"variant1"
```
mattyg, line 108:
```suggestion
    "MyEnumType": "Variant1"
```

zippy (approved):
This looks great apart from suggestions others have left.

ThetaSinner (changes requested):
I wouldn't know how to upgrade my Holonix, or NPM and Rust dependencies based on this guide. I'd also like some more details on Rust code changes required please.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
pdaoust: I'm not too worried about non-breaking changes -- unless this *is* a breaking change that didn't get put into the changelog?

src/pages/get-started/install-advanced.md
pdaoust:
```suggestion
If you want to enable or disable certain Holochain features, such as unstable features, it's best to do this in a local flake file. [Read the 'Customized Holochain build'](https://github.com/holochain/holonix?tab=readme-ov-file#customized-holochain-build) on the Holonix readme to find out how. Keep in mind that, because you'll be creating a custom Holochain binary, you won't be able to take advantage of the package cache, so it'll take a while to compile Holochain on your machine or run CI tasks with it.
```

pdaoust:
> I wouldn't know how to upgrade my Holonix, or NPM and Rust dependencies based on this guide. I'd also like some more details on Rust code changes required please.

@ThetaSinner I've rewritten the guide; is this more what you're thinking now?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+
+Holochain 0.4 introduces a new flow for app installation that lets an agent supply a membrane proof later with a new [`ProvideMemproofs` endpoint](https://docs.rs/holochain_conductor_api/0.4.0-rc.2/holochain_conductor_api/enum.AppRequest.html#variant.ProvideMemproofs). This allows them to get their freshly generated agent key, submit it to some sort of membrane proof service, and get a membrane proof.
+
+Because of this, after the membrane proof has been provided, the [`DisabledAppReason`](https://docs.rs/holochain_types/0.4.0-rc.2/holochain_types/app/enum.DisabledAppReason.html) enum, used in the [`AppInfo`](https://docs.rs/holochain_conductor_api/0.4.0-rc.2/holochain_conductor_api/struct.AppInfo.html) response, will be a new `NotStartedAfterProvidingMemproofs` variant until the app is started.
+
+The only change you should need to make to existing code is to make sure you're handling this new variant in your `match` blocks (Rust) or `switch` blocks (JavaScript).
+
+### Change in enum serialization
+
+The default serialization for bare (no-data) enum variants has changed. Previously, they would look like this (JSON representation):
```
jost-s, line 395:
```suggestion
The default serialization for unit-like enum variants has changed. Previously, they would look like this (JSON representation):
```

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+     // some dependencies
+-    "@holochain/client": "^0.17.1",
+-    "@holochain/tryorama": "^0.16.0",
++    "@holochain/client": "^0.18.0-rc.1",
++    "@holochain/tryorama": "^0.17.0-rc.0",
+     // more dependencies
+   },
+```
+
+Then navigate to the `tests/` folder and run your package manager's update command like you did at the project's top level in the last step.
```
jost-s, line 116: `npm update` should only be run from the project's root folder.

pdaoust: awesome, I wasn't sure about that

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+
+```diff:json
+   "dependencies": {
+-    "@holochain/client": "^0.17.1",
++    "@holochain/client": "^0.18.0-rc.1",
+     // more dependencies
+   },
+```
+
+Then run navigate to the `ui/` folder and run your package manager's update command.
```
jost-s: You can update all dependencies in root, test and ui folders and then run `npm i` just once.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Rust
+
+Update the `hdk` and `hdi` version strings in the project's root `Cargo.toml` file:
+
+```diff:toml
+ [workspace.dependencies]
+-hdi = "=0.4.6"
+-hdk = "=0.3.6"
++hdi = "=0.5.0-rc.1"
```
pdaoust, line 48: These all need to be fixed once a release is cut. There are many of these.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+
+Then run navigate to the `ui/` folder and run your package manager's update command.
+
+## Update your application code
+
+Here are all the breaking changes you need to know about in order to update your app for Holochain 0.4.
+
+### Unstable features removed by default
+
+The biggest change for 0.4 is that some unstable features aren't compiled into official Holochain releases by default. We're doing this to focus on a stable core Holochain with the minimal feature set that most projects need. The removed features are still available as Rust feature flags, so you can compile a custom Holochain binary if you need them. Here's the full list, along with their feature flags:
```
jost-s, line 134:
```suggestion
The biggest change for 0.4 is that some features are marked `unstable` and arent' compiled into Holochain by default. We're doing this to focus on a stable core Holochain with the minimal feature set that most projects need. The removed features are still available as Rust feature flags, so you can compile a custom Holochain binary if you need them. Here's the full list, along with their feature flags:
```

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+    * Related to DPKI (enable `unstable-dpki` if you want to use these, which are both new to 0.4):
+        * [`get_agent_key_lineage`](https://github.com/holochain/holochain/blob/holochain-0.4.0-rc.2/crates/hdk/src/agent.rs#L10-L20)
+    * Related to block lists (allowing peers in a hApp to selectively block others who haven't necessarily produced invalid data or forked their source chain):
+        * `block_agent`
+        * `unblock_agent`
+    * Other:
+        * [`schedule`](https://github.com/holochain/holochain/blob/holochain-0.4.0-rc.2/crates/hdk/src/time.rs#L67-L137)
+        * An unimplemented `sleep` function has been removed completely
+
+`unstable-functions` is a flag used by both the Holochain conductor _and_ the [`hdi`](https://docs.rs/hdi/latest/hdi) and [`hdk`](https://docs.rs/hdk/latest/hdk) crates, and some of the functions also need other Holochain features enabled (e.g., `is_same_agent` requires a conductor with `unstable-dpki` enabled; see the list above).
```
jost-s, line 153: I think that's an implementation detail that app devs don't need to know about.
```suggestion
Some of the unstable functions need other Holochain features enabled (e.g., `is_same_agent` requires a conductor with `unstable-dpki` enabled; see the list above).
```

pdaoust: It seems like those few people who need to know about this will need to know that they have to turn on `unstable-functions` in both the zomes and the custom holochain binary they build, along with the feature they depend on, in order for unstable functions to be plumbed all the way through. Is that not true?

jost-s: Okay yes, then I'd be explicit about it and say "`unstable-functions` has to be enabled both when building the Holochain binary **and** in the Cargo.toml of zomes using `hdi` and `hdk`.

pdaoust: oh -- I tried to say that :) I'll go back and see how I can be more explicit.

src/pages/get-started/install-advanced.md
ThetaSinner: "to compile Holochain on your machine. The same will be true on CI, where you should consider setting up your own caching"

Something like that?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+
+This section is marked as optional because it's possible that new dependencies could introduce new issues as well as fixing existing conflicts or problems. To make it possible to roll back this change, it might be a good idea to commit the changes you've made so far to source control. Then you can run:
+
+```shell
+cargo update
+```
+
+This will update your `Cargo.lock` with the latest versions of all libraries that the constraints in your `Cargo.toml` files will allow. Now you should try building your project again to see if that has resolved your issue.
+
+### JavaScript tooling
```
ThetaSinner, line 71: Confusing title, tryorama is also javascript tooling. Maybe something like "Node dependencies in the project root`?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+npm update
+```
+```shell
+pnpm update
+```
+```shell
+yarn upgrade
+```
+
+### Tryorama tests
```
ThetaSinner, line 101: Inaccurate, we're updating tryorama, not the tests. Maybe `Node dependencies for the test workspace`?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+bun update
+```
+```shell
+npm update
+```
+```shell
+pnpm update
+```
+```shell
+yarn upgrade
```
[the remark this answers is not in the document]
ThetaSinner: I'd also move this to the end of the section because we may as well update all our dependencies at once once we're done editing `package.json` files. It can all be done in one command.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+Read the [Holonix readme](https://github.com/holochain/holonix?tab=readme-ov-file#customized-holochain-build) to find out how to enter a development shell with a custom-compiled Holochain build with these flags enabled.
+
+If you're using Holochain as a library in your project, edit your `Cargo.toml` file's `holochain` dependency to enable the flags. Here's an example that enables countersigning:
+
+```diff:toml
+ [dependencies]
+-holochain = { version = "0.3.6" }
+ # "unstable-functions" is needed in order for coordinator zomes to call the
+ # host functions related to countersigning.
++holochain = { version = "0.4.0-rc.2", features = ["unstable-countersigning", "unstable-functions"] }
```
ThetaSinner, line 171: Nobody is using countersigning. It's so complex that I don't believe anybody has figured out how to use it and put it in their app without talking to us. It took me weeks, full time, trying to figure out how to get it behaving properly and then it was packed full of bugs. The chances that anybody else has bothered going through that is low.

Anybody building Holochain into the project to make their own runtime, should know how to use Rust features. I'd just drop this section. The section above that describes how you need to build Holochain with feature flags and also the HDK/HDI should be sufficient I believe.

ThetaSinner: Part of the reason for disabling this stuff is that most people aren't using what we're putting behind feature flags. We wanted to let people use it if they wanted to, but the main reason for using Rust features really was to be able to keep testing that code until we can get to it and make it work properly.

ThetaSinner (commented):
Pausing again because I got stuck

An app scaffolded at 0.3 doesn't `npm start` because the UI is broken

```
[0]     48 |     await this.fetchWorzle();
[0]     49 |     toRaw(this.client).on("signal", signal => {
[0]   > 50 |       if (signal.zome_name !== "ohfour") return;
[0]        |                  ^^^^^^^^^
[0]     51 |       const payload = signal.payload as OhfourSignal;
[0]     52 |       if (payload.type !== "EntryCreated") return;
[0]     53 |       if (payload.app_entry.type !== "Worzle") return;
```

I'm using Vue but I'm pretty sure this issue should show up for an app scaffolded with any UI framework.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
pdaoust: Would it work to just change it to e.g. `hdi = "^0.4.6"`?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
pdaoust: I do know of one project in the wild that's using countersigning, but point taken -- those who need the features will probably have the chops to do what I'm describing.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,434 @@
...
+
+### JavaScript tooling
+
+If you've created your hApp using our scaffolding tool, you should be able to follow these instructions. If you've created your own project folder layout, adapt these instructions to fit.
+
+Edit your project's root `package.json` file to update the developer tools:
+
+```diff:json
+   "devDependencies": {
+     "@holochain-playground/cli": "^0.300.1",
```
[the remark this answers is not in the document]
pdaoust: Playground 0.300 seems to work fine with hApps built for 0.4 🤷🏼‍♂️

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,417 @@
...
+   app.disable_clone_cell(disable_payload).await?;
+   let delete_payload = DeleteCloneCellPayload {
+-    clone_cell_id: CloneCellId::CellId(cloned_cell.cell_id),
++    clone_cell_id: CloneCellId::DnaHash(cloned_cell.cell_id.dna_hash()),
+   };
+   app.delete_clone_cell(delete_payload).await?
+ }
+```
+
+### JavaScript client now receives system signals
```
pdaoust, line 319: @ThetaSinner I've added a bit of JS to update client/Tryorama code for the breaking change in the signal handler. Does this look good?

src/pages/get-started/install-advanced.md
pdaoust: wait! I just discovered this is already fixed in `main`. Guess I just didn't sync this branch with `main` lately.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
pdaoust: @ThetaSinner I'd like to see the code that's not building for you -- it's working for me. Looks like you're trying to read `SignalType.AppSignal` rather than `SignalType.App`?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,417 @@
...
+## Quick instructions
+
+To upgrade your hApp written for Holochain 0.3, follow these steps:
+
+1. Check whether your `flake.nix` contains `github:holochain/holonix`. If it does, then you can continue to the next step. Otherwise it will contain `github:holochain/holochain`. If so, then follow the [Holonix upgrade guide](/resources/upgrade/upgrade-new-holonix/) to update to the newest Holonix command-line developer environment (we're not providing a 0.4 version of Holochain through the old Holonix).
+2. Update your `flake.nix` to use the 0.4 version of Holochain by changing the version number in the line `holonix.url = "github:holochain/holonix?ref=main-0.3"` from 0.3 to 0.4. This will take effect later when you enter a new Nix shell. It's important to update your Nix flake lock at this point, to ensure you benefit from the cache we provide:
+
+    ```shell
+    nix flake update
+    nix develop
```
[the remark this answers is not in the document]
pdaoust: hm, a bug in my shell styling. Should be a one-liner for easy copy/pasting anyway.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,413 @@
...
+
+### `CloneCellId` changes
+
+The `CloneCellId::CellId` enum variant has become [`DnaHash`](https://docs.rs/holochain_zome_types/0.4.0-rc.1/holochain_zome_types/clone/enum.CloneCellId.html) and contains, naturally, a `DnaHash` value. This type is used when enabling, disabling, or deleting clones from the app API or a coordinator zome.
+
+#### In coordinator zomes
+
+Edit any coordinator zome code that uses functions from `hdk::clone`:
+
+```diff:rust
```
pdaoust, line 161: Code compiles, functions not tested.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,413 @@
...
+   };
+   delete_clone_cell(delete_input)
+ }
+```
+
+#### In JavaScript front-end
+
+Edit any client code that manipulates cloned cells:
+
+```diff:typescript
```
pdaoust, line 198: `tsc` can compile this too -- again not tested.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+
+### Unstable features removed by default
+
+The biggest change for 0.4 is that some features are marked `unstable` and aren't compiled into Holochain by default. We're doing this to focus on a stable core Holochain with the minimal feature set that most projects need. The removed features are still available as Rust feature flags, so you can compile a custom Holochain binary if you need them. Here's the full list, along with their feature flags:
+
+* **[Countersigning](/concepts/10_countersigning/) `unstable-countersigning`**: Allows agents to temporarily lock their chains in order to agree to write the same entry to their respective chains.
+* **DPKI / DeepKey `unstable-dpki`**: Implements a distributed public key infrastructure, allowing agents to manage their keys and confirm continuity of their identity across devices and source chains. Read the [DeepKey readme](https://github.com/holochain/deepkey) for more info.
+* **[DHT sharding](/concepts/4_dht/) `unstable-sharding`**: Lessens the load of each peer in large network, allowing them to take responsibility for validating and storing only a subset of a DHT's entire data set.
```
ThetaSinner, line 131: This feature flag doesn't remove code, it just forces configuration. By default in 0.4 you must set the network tuning parameter `gossip_arc_clamping` in your conductor config to "empty" or "full". The default is "full" and anything else will produce an error. If you enable the feature then you can still set "empty" or "full" but an extra option is available, "none". Where "none" is the default and let's Kitsune manage arc sizing.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+* **[Warrants](/concepts/7_validation#remembering-validation-results) `unstable-warrants`**: Spreads news of validation failure around the network, allowing peers to recognize bad actors and take action against them, even if they haven't directly interacted with them.
```
ThetaSinner, line 132: This feature is new and we know there are bugs, use with caution

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+* **Chain head coordination `chc`**: Developed for Holo, allows source chain changes to be copied from one device to another without causing chain forks. (This pre-existing feature flag is no longer enabled by default.)
```
ThetaSinner, line 133: Technically it allows you to back up your source chain writes to a remote server that implements the CHC protocol over HTTP. Then you can restore your source chain from that backup. Being able to move machines is a side-effect of what the feature does.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+* **HDI/HDK functions `unstable-functions`**:
+    * Related to countersigning (enable `unstable-countersigning` if you want to use these):
+        * [`accept_countersigning_preflight_request`](https://github.com/holochain/holochain/blob/holochain-0.4.0-rc.2/crates/hdk/src/countersigning.rs#L3-L22)
+    * Related to DPKI (enable `unstable-dpki` if you want to use these, which are both new to 0.4):
```
ThetaSinner, line 137: Both? Is there a function missing here then? Should this also list `is_same_agent`?

src/pages/resources/upgrade/upgrade-holochain-0.4.md
ThetaSinner: This is just a matter of running the app and seeing whether created entries show up immediately on the UI or whether you have to reload the page. If they show up right away, this is working

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
++    clone_cell_id: cell_id[0],
+   });
+ }
+```
+
+If you're writing an application that uses the admin API, `AdminClient#deleteCloneCell` changes in the same way as `enableCloneCell` and `disableCloneCell`.
+
+### JavaScript client now receives system signals
+
+For JavaScript front ends and Tryorama tests, the signal handler callback for `AppWebsocket.prototype.on("signal", cb)` should now take a [`Signal`](https://github.com/holochain/holochain-client-js/blob/main-0.4/docs/client.signal.md). Update your code to look like this:
```
ThetaSinner, line 222: Can note here that system signals are currently only used for countersigning, which is disabled. So you can filter out system signals for now.

We are planning to use these for other things so it's probably worth making it clear that this is explicitly filtering them out.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+If you're one of the rare folks who have been using these two actions, the structs have changed. Take a look at the Rustdoc and update your code accordingly:
+
+* `CloseChain::new_dna_hash` has been replaced with [`new_target`](https://docs.rs/holochain_zome_types/0.4.0-rc.1/holochain_zome_types/action/struct.CloseChain.html), which has a type of `Option<MigrationTarget>`. [This new type](https://docs.rs/holochain_zome_types/0.4.0-rc.1/holochain_zome_types/prelude/enum.MigrationTarget.html) is an enum of `Dna(DnaHash)` or `Agent(AgentHash)`.
+
+* [`OpenChain::prev_dna_hash`](https://docs.rs/holochain_zome_types/0.4.0-rc.1/holochain_zome_types/action/struct.OpenChain.html) has been changed to match; its type is a non-optional `MigrationTarget`. It also gets a new field, `close_hash`, which is the `ActionHash` of the corresponding `CloseChain` action.
+
+### `InstallApp` agent key is optional
+
+!!! info
+This change is only relevant if you're building a runtime that can install hApp bundles.
```
ThetaSinner, line 268: If you're doing that, you might just as well be using Holochain as a library and need a whole other list of changes. I'd just say "using the Rust client to install apps on Holochain" or something like that here

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+
+The `CountersigningSuccess` signal sent to the client now gives the action hash of the successful countersigned commit rather than its entry hash. In your client code, whether JavaScript or Rust, update your signal handler to deal with a `HoloHash<Action>` rather than a `HoloHash<Entry>` when it receives this kind of signal.
+
+## Subtle changes
+
+### [DHT sharding](/concepts/4_dht/) is disabled by default
+
+This feature needs more performance and correctness testing before it's production-ready, but it can be turned on by [compiling Holochain with the `unstable-sharding` flag enabled](#unstable-features-removed-by-default).
+
+It's unknown exactly what might happen if nodes with DHT sharding disabled try to gossip in the same network as nodes without DHT sharding. Presumably this is still possible, but it might cause unexpected behaviors!
```
ThetaSinner, line 348: No, that is known. Because we've run with `gossip_arc_clamping = "full"` in the same network are peers with dynamic arcs before. It will work fine but the peers who have chosen to enable `unstable-sharding` and left their `gossip_arc_clamping = "none"` might run into bugs with arc sizing. That's the reason we disabled it and until we integrate Kitsune2 and re-enable sharding, we're not going to provide support for people who run into bugs with this. At least not beyond helping diagnose that this was indeed the issue.

src/pages/resources/upgrade/upgrade-holochain-0.4.md
```diff
@@ -0,0 +1,360 @@
...
+
+### Dynamic database encryption, folder structure changes
+
+Previous conductor and keystore SQLite databases used a hardcoded encryption key; v0.4 now uses a dynamic key. Additionally, the database paths have changed. These two things mean you won't be able to import data from v0.3.
```
ThetaSinner, line 352: There is an automatic migration of paths. If databases are detected at the old paths then Holochain will move them. We didn't provide a "re-encrypt" migration from the hard-coded key to dynamic keys though. It can be done, it's not impossible but we didn't think the value was high enough to justify the time.

So I'd word this to say that it's the encryption that prevents data being imported and if people want to keep their data they either need to export it through their app's UI and re-import on 0.4 or they need to provide their users with tooling to migrate the encryption keys.

Maybe that's a bit too much detail really, but conductor migration is something I'd like us to be getting better at over time :)

Issue #510, opened 2025-01-09
App structure 1: zomes


pdaoust:

pdaoust:
Closed by #511 and #512.

PR #511, opened 2025-01-15
Zomes page

pdaoust:
Adds a page that talks about thinking about, structuring, and creating a zome. Tells what goes into a zome, etc.


pdaoust:
Closes #510 .

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+!!!
+
+A zome has access to Holochain's host API and also exposes functions of its own. Some of these functions are **required callbacks** and some of them you invent yourself to create your back end's API.
+
+There are two kinds of zome:
+
+* An **integrity zome** defines a set of data types --- your application's schema --- and validation rules for operations that create, update, or delete data of those types. If you're used to thinking of apps in terms of [model-view-controller (MVC)](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller), it's like your application's model.
+* A **coordinator zome** defines a set of functions for interacting with data and peers. You can think of it as your application's controller.
+
+Zomes are usually created as pairs --- an integrity zome that defines a portion of your schema and a coordinator zome that defines <abbr title="create, read, update, delete">CRUD</abbr> functions for that schema. You don't have to do it this way though; coordinator zomes don't need an integrity zome if they don't manipulate data, or they can depend on multiple integrity zomes, or multiple coordinators can depend on the same integrity zome.
```
ThetaSinner, line 42:
```suggestion
Zomes are usually created as pairs --- an integrity zome that defines a portion of your data types and rules, and a coordinator zome that defines functions that operate on those data types. You don't have to do it this way though; coordinator zomes don't need an integrity zome if they don't manipulate data, or they can depend on multiple integrity zomes, or multiple coordinators can depend on the same integrity zome.
```

A schema suggests static rules defined in structured data to me. We usually talk about "rules of the game" with integrity zomes and I think "rules" better suggest that you define your own rules in code.

For coordinators, you are not limited to defining simple CRUD against the data. You could write your coordinator that way, but I think it's a mistake to repeatedly cross the WASM boundary for CRUD and have all your logic on the front end. I think it makes far more sense to have your backend doing more creative work. Like create_entry + create_link, is already doing more than a "C" and the two operations will succeed or fail together. That's quite important.

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+If you mean for your zomes to be reused by other projects, you can share them via a public Git repository or [crates.io](https://crates.io) (tag your crates with `#holochain` so others can find them).
```
ThetaSinner, line 44: Is anybody using `crates.io`? I haven't come across that yet. I'm assuming there's some reason people are using `git` dependencies. Any insight here @matthme or @zippy?

src/pages/build/application-structure.md
zippy, line 42:
```suggestion
Zomes are usually created as pairs --- an integrity zome that defines your schema and validations rules, along with a coordinator zome that defines <abbr title="create, read, update, delete">CRUD</abbr> functions for that schema. You don't have to do it this way though; coordinator zomes don't need an integrity zome if they don't manipulate data, or they can depend on multiple integrity zomes, or multiple coordinators can depend on the same integrity zome.
```

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+[Read more on the Zomes page](/build/zomes/).
+
+### DNA
+
+One or more zomes are bundled into a **DNA**. When two or more participants install and run a DNA, an isolated peer-to-peer network is created among them to interact and store shared data.
```
ThetaSinner, line 50:
```suggestion
One or more zomes are bundled into a **DNA**, including at least one integrity zome. When two or more participants install and run a DNA, a new peer-to-peer network is created among them to interact and store shared data.
```

src/pages/_data/navigation/mainNav.json5
```diff
@@ -25,6 +25,9 @@
       ]
     },
     { title: "Build", url: "/build/", children: [
+        { title: "Application Structure", url: "/build/application-structure/", children: [
+          { title: "Zomes", url: "/build/zomes/" },
```
zippy, line 29: I notice when I click on the zomes page, or any of the other sub-items (identifiers, etries,..) over in the right hand navigation that the hierarchy gets closed back down to just `Build` which is very confusing.  Can we fix that?

src/pages/build/application-structure.md
ThetaSinner: An "isolated" network has an existing meaning that we do not meet https://www.meter.com/resources/what-is-network-isolation

zippy, line 50: Here you use the term "Particpant" elsewhere I see "Peer".  This should be consistent.

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+**A DNA, and the network created for it, is uniquely defined by its integrity zomes, plus any modifiers.** The hash of the integrity zomes plus modifiers is called the **DNA hash**, and is the unique identifier for the network.
+
+!!! info Why not coordinator zomes?
+
+Coordinator zomes are bundled with a DNA, but they don't contribute to its uniqueness. This lets you hot-swap coordinators as you fix bugs and add features, without causing the DNA to **fork** a new network. The only things that should cause a fork are changes to integrity code --- the 'rules of the game' for the participants.
```
ThetaSinner, line 56: This is good but I think we want to talk about forking separately. That's an option, but the concept of migration is different to a fork. I don't know what the best thing to say here would be because you currently can't do DNA migration.

Maybe we should be direct and say that for now the lifecycle of data is completely tied to the DNA until we implement migrations?

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+!!!
+
+Because each DNA has its own isolated peer network and data store, you can use the DNA concept to come up with creative approaches to privacy and access, separation of responsibilities, or data retention.
```
ThetaSinner, line 60: There is no privacy without encryption, DHT data is not protected by anything Holochain does. Its *integrity* is protected, but read access to it is not

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+::: intro
+A **zome** (short for chromosome) is a module of executable code within a [**DNA**](/resources/glossary/#dna). It's the smallest unit of modularity in a Holochain application.
+:::
+
+## How a zome is structured
+
+A zome is just a [WebAssembly module](https://webassembly.github.io/spec/core/syntax/modules.html) that exposes public functions. The **conductor** (the Holochain runtime) knows about these functions and calls them at different points in the application's lifetime. Some functions have special names and are called **lifecycle callbacks**<!-- TODO uncomment when lifecycle callbacks PR is merged [lifecycle callbacks](/build/lifecycle-events-and-callbacks/)-->, and you can also define arbitrarily named functions of your own to serve as your zome's API.
+
+For Rust developers, we've created an SDK called the [Holochain Development Kit (HDK)](https://crates.io/crates/hdk/). It lets you define functions, exchange data with the host (the Holochain VM), and access all of the host's functionality.
```
zippy, line 13: This feels too confusing here and out of context, I suggest just deleting it, or somehow relating it zomes.  i.e. it's not actually part of "how a zome is structured" it's info about writing zome code.

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+### hApp
+
+Multiple DNAs come together in a **hApp** (Holochain app). Each DNA fills a named **role** in the hApp, and can be seen as a [microservice](https://en.wikipedia.org/wiki/Microservices).
```
ThetaSinner, line 64: That's a nice analogy, I like that

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+Each peer generates its own public/private key pair when they install a hApp. Their public key acts as their **agent ID** which identifies them as a participant in all the networks created for all the DNAs in the hApp. When a DNA is activated, it's bound to this key pair and becomes a DNA instance or **cell**.
+
+The hApp can specify a few provisioning strategies for its DNAs:
+
+* A cell can be instantiated at app installation time.
+* A new cell can be **cloned** from an existing DNA at any time _after the hApp is installed_, with an optional limit on the number of clones.
+
+A hApp can optionally include a web-based UI that supporting Holochain runtimes <!-- TODO: link --> can serve to the user.
```
ThetaSinner, line 73: I think `runtimes` deserves its own introduction/definition. I think this term is not accurate to the standard meaning of the term and is something we've fallen into using internally. I'd probably put it in bold like the other Holochain defined terms and give it a quick explanation

src/pages/build/application-structure.md
```diff
@@ -0,0 +1,87 @@
...
+
+!!! info A hApp always runs locally
+
+The big difference with peer-to-peer stacks like Holochain is that **all the code** --- both the back end and the front end --- **runs on the devices of the participants themselves**.
+
+That means that a DNA doesn't exist as some piece of code that runs 'out there somewhere' --- instead it runs from the perspective of an individual. DNA + agent = cell.
+
+There can still be bots or system-level services that do automated tasks. Those functions just have to be handled by one of the peers, and that peer doesn't have to be a human.
```
ThetaSinner, line 81: Might be nice to tie back to the "agent" term here? There's a reason we use agent to cover the actor and it's this

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+title: "Zomes"
+---
+
+::: intro
+A **zome** (short for chromosome) is a module of executable code within a [**DNA**](/resources/glossary/#dna). It's the smallest unit of modularity in a Holochain application.
+:::
+
+## How a zome is structured
+
+A zome is just a [WebAssembly module](https://webassembly.github.io/spec/core/syntax/modules.html) that exposes public functions. The **conductor** (the Holochain runtime) knows about these functions and calls them at different points in the application's lifetime. Some functions have special names and are called **lifecycle callbacks**<!-- TODO uncomment when lifecycle callbacks PR is merged [lifecycle callbacks](/build/lifecycle-events-and-callbacks/)-->, and you can also define arbitrarily named functions of your own to serve as your zome's API.
```
ThetaSinner, line 11: It discovers the functions rather than knowing about them. That's really only for callbacks though I think and it's up to the end user to know about zome function names to be able to call them

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+## The two types of zomes
+
+### Integrity
+
+An **integrity zome** defines a portion of your application's data model or schema. This includes not just the structure of the data but also _validation rules for operations that manipulate this data_.
+
+!!! info Keep your integrity zomes small
+
+While you can define arbitrary public zome functions in your integrity zome, in practice it makes fixing bugs and adding features difficult, because every code change to an integrity zome [modifies the hash of the DNA](/build/application-structure/#dna) and causes a fork of the database. So it's better for maintainability if the integrity zome _only_ contains your data model and the necessary callbacks that inform Holochain about that model.
```
zippy, line 23: Can we reverse this?  I.e. instead of telling people what they could possibly do up front, tell them what they should do, and then you might have a caveat about what they could do if they have to for example:

```
Your integrity zome should only contain your data model and the necessary callbacks that inform Holochain about that model.  Although you can define arbitrary public zome functions in your integrity zome, in practice it makes fixing bugs and adding features difficult, because every code change to an integrity zome [modifies the hash of the DNA](http://localhost:8080/build/application-structure/#dna) and causes a network fork.
```

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+Finally, your integrity zome defines validation callbacks <!-- TODO: uncomment once lifecycle events PR is merged [validation callbacks](/build/lifecycle-events-and-callbacks/#define-a-validate-callback)--> that check for correctness of data and actions. Holochain runs this on an agent's own device when they attempt to author data, and on other peers' devices when they're asked to store and serve data authored by others.
+
+#### Create an integrity zome
+
+**The easy way to create an integrity zome** is to [scaffold a new hApp](/get-started/3-forum-app-tutorial/). The scaffolding tool will generate all the project files, including scripts to test and build distributable packages, and it can also scaffold boilerplate code for all your app's required callbacks and data types.
+
+If you want to start from scratch, first make sure you have Rust, Cargo, and the `wasm32-unknown-unknown` Rust toolchain installed on your computer. Then create a library crate:
```
zippy, line 39:
```suggestion
If you want to start from scratch without the scaffolding, first make sure you have Rust, Cargo, and the `wasm32-unknown-unknown` Rust toolchain installed on your computer. Then create a library crate:
```

src/pages/build/zomes.md
ThetaSinner: For "lifecycle callbacks", that only covers things like `init` and migration hooks etc. Things like `validate` and `post_commit` are event driven but not by the lifecycle of the app. I'd prefer to just call them "callbacks" and optionally define "lifecycle hooks" to cover a class of callbacks

src/pages/build/zomes.md
ThetaSinner, line 13:
```suggestion
For Rust developers, we've created an SDK called the [Holochain Development Kit (HDK)](https://crates.io/crates/hdk/). It lets you define functions, exchange data with the host (the Holochain runtime), and access all of the host's functionality.
```

zippy (changes requested):
This is great progress, let's keep iterating like this!  

Minor change requests made.

src/pages/build/application-structure.md
zippy: I deleted the "portion" here because it's confusing in this context.  maybe it could come back but it would need  "your app's schema"

src/pages/build/zomes.md
ThetaSinner, line 23: Solid advice, objecting again to "fork", we have no mechanism to fork and we're mostly thinking about migration.

Sure you could use a migration to use a "fork" but it's not the thing we're focused on enabling with allowing people to update integrity code

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Integrity
+
+An **integrity zome** defines a portion of your application's data model or schema. This includes not just the structure of the data but also _validation rules for operations that manipulate this data_.
+
+!!! info Keep your integrity zomes small
+
+While you can define arbitrary public zome functions in your integrity zome, in practice it makes fixing bugs and adding features difficult, because every code change to an integrity zome [modifies the hash of the DNA](/build/application-structure/#dna) and causes a fork of the database. So it's better for maintainability if the integrity zome _only_ contains your data model and the necessary callbacks that inform Holochain about that model.
+
+Because these callbacks only need a small portion of Holochain's functionality, you don't need the entire `hdk` crate to write an integrity zome. Use the smaller [`hdi`](https://crates.io/crates/hdi) crate instead.
```
ThetaSinner, line 25: This should be elevated. I think let's not confuse things by letting people know that it's possible to use the HDK in an integrity zome, let's just stick with saying that the HDI is for integrity zomes.

If you call HDK functions in a validation context, it will just fail. 

The reason for having them separate isn't as simple as size, it's about stability guarantees. For the same reason that you don't want to be updating your zome code too often like you've discussed above, you don't want to be forced to upgrade to a new HDK as that changes. The intention is to make changes to the HDI really rarely as Holochain becomes stable.

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+Your integrity zome tells Holochain about the types of [entries](/build/entries/) and [links](/build/links-paths-and-anchors/) it defines with macros called [`hdk_entry_types`](https://docs.rs/hdi/latest/hdi/attr.hdk_entry_types.html) and [`hdk_link_types`](https://docs.rs/hdi/latest/hdi/attr.hdk_link_types.html) added to enums of all the entry and link types. These create lifecycle callbacks that are run at DNA install time and give Holochain the info it needs. Read more in [Define an entry type](/build/entries/#define-an-entry-type) and [Define a link type](/build/links-paths-and-anchors/#define-a-link-type).
+
+Finally, your integrity zome defines validation callbacks <!-- TODO: uncomment once lifecycle events PR is merged [validation callbacks](/build/lifecycle-events-and-callbacks/#define-a-validate-callback)--> that check for correctness of data and actions. Holochain runs this on an agent's own device when they attempt to author data, and on other peers' devices when they're asked to store and serve data authored by others.
```
ThetaSinner, line 33:
```suggestion
Finally, your integrity zome defines validation callbacks <!-- TODO: uncomment once lifecycle events PR is merged [validation callbacks](/build/lifecycle-events-and-callbacks/#define-a-validate-callback)--> that check for correctness of data and actions. Holochain runs this on an agent's own device when they attempt to author data, and when they're asked to store and serve data authored by others.
```

src/pages/build/zomes.md
ThetaSinner, line 39: `wasm32-unknown-unknown` is a compiler target, not a toolchain -> https://github.com/holochain/holochain/blob/develop/rust-toolchain.toml

It's either added with a `rust-toolchain.toml` or with `rustup target add wasm32-unknown-unknown`

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+Then add some necessary dependencies to your new `Cargo.toml` file:
+
+```diff:toml
+ [dependencies]
++hdi = "=0.5.0-rc.1"
++serde = "1.0"
+```
+
+At the very minimum, make sure your code exposes a `validate` callback <!-- TODO: uncomment once lifecycle events PR is merged [`validate` callback](/build/lifecycle-events-and-callbacks/#define-a-validate-callback)--> and [defines some entry types](/build/entries/#define-an-entry-type).
```
ThetaSinner, line 53: Is that required? I don't think it is, I think it's optional to define either types or a validate callback. Your zome isn't very useful without one or the other. I'd soften the wording here though

```suggestion
At this point your can define a `validate` callback <!-- TODO: uncomment once lifecycle events PR is merged [`validate` callback](/build/lifecycle-events-and-callbacks/#define-a-validate-callback)--> and [define some entry types](/build/entries/#define-an-entry-type).
```

ThetaSinner: I'm pretty sure Holochain lets you define types without a `validate` callback. Maybe we should prevent that, but I'm not aware that we do at the moment.

I think I may have said differently recently, maybe we should check this

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+-hdi = "=0.5.0-rc.1"
++hdk = "=0.4.0-rc.1"
+ serde = "1.0"
+```
+
+And there aren't any required callbacks to define in your code.
+
+## Define a function
+
+The conductor and the zome pass function input and output data to each other by first allocating the data in a piece of memory that they both share, then passing a pointer to the location and size of the data. This is rather complicated, so we've built a macro called [`hdk_extern`](https://docs.rs/hdk/latest/hdk/prelude/attr.hdk_extern.html) that makes it as easy as defining a `pub fn`.
```
ThetaSinner, line 80: I wonder how much value this is adding to include these details. I think something along the lines of "we've provided a macro to simplify the integration with WASM".

To dip into a level of detail that most Rust programmers will get, you could say something like "it's a bit like the `extern "C"` that you can use to make your Rust code callable from C, here you're making your function callable by Holochain"

src/pages/build/zomes.md
```diff
@@ -0,0 +1,144 @@
...
+
+You can also explicitly return an error with the [`wasm_error`](https://docs.rs/hdi/latest/hdi/prelude/macro.wasm_error.html) macro:
+
+```rust
+use hdk::prelude::{hdk_extern, ExternResult, wasm_error};
+
+#[hdk_extern]
+pub fn check_age_for_18a_movie(age: u32) -> ExternResult<()> {
+    if age >= 18 {
+        Ok(())
```
ThetaSinner, line 123:
```suggestion
        return Ok(());
```

Has this code been built? `cargo build` should have picked that up

ThetaSinner (commented):
Quite a few queries and suggestions but I think this is a good overview of the pieces of a hApp

src/pages/build/zomes.md
pdaoust: not yet; I remembered I hadn't done that yet and was considering putting this back into draft mode. Sorry I missed that!

src/pages/build/application-structure.md
pdaoust: oh, that's kind of what I was glomming into the definition of 'CRUD' -- the creative stuff. Ditto re: schema -- I see it as an imperative schema where devs are used to writing declarative schemas. But what I'm getting here is that these words are conveying things that aren't accurate, so I'll accept the suggestion.

src/pages/build/application-structure.md
pdaoust: it'd be good to pave the cowpaths, so I'll defer to what's actually being done in practice.

src/pages/build/application-structure.md
pdaoust: On a casual reading it seems like it meets the meaning perfectly -- what nuance am I missing?

pdaoust: I've been using different terms for differnt situations:

* 'Participant' as the general hand-wavey term for user/agent/whatever
* 'Peer' as a 'someone else' in relation to a primary participant in an interaction -- the Bob to an Alice
* 'Agent' as the technical term for someone/thing in control of a cell
* 'User' when I can't think of a better term for 'human operator of a cell'

How do those feel? And have I been using them inconsistently? Sometimes it's a struggle to hit on the best term in a given context. I could also just define 'agent' at some root-level documentation and use it consistently everywhere, although sometimes the more specific term really is the best one.

src/pages/build/zomes.md
pdaoust: oh good point about fewer upstream changes.

src/pages/build/zomes.md
pdaoust: I'd love a better word -- any ideas? We've used 'fork' forever to describe creating a new network space w/o carrying part of the old data and peer topology with it, but I can see how it could imply those things. I grant that 'fork' does suggest a contiguous history that splits in two.

src/pages/build/zomes.md
pdaoust: okay! noted

src/pages/build/zomes.md
pdaoust: I created a new section for it that gives the (somewhat missing) piece of informtaion that a zome is just a library crate. (In practice anyway)

src/pages/_data/navigation/mainNav.json5
pdaoust: Attempting to fix in #513 , which is still a draft.

src/pages/build/application-structure.md
zippy: Joel has used it for utilities crates:  https://crates.io/crates/hc_iz_membrane_manager

src/pages/build/application-structure.md
zippy: I think we should just define this by example in our case and point to various runtimes (whenever we actually add this).

PR #513, opened 2025-01-17
try to fix issue with deep navs collapsing when you navigate to a page within them

pdaoust:
The sidebar nav on a page that's three levels deep looks like this:

image

It ought to look like this:

image


pdaoust:
@zippy take a look at https://deploy-preview-513--developer-portal-production.netlify.app/build/identifiers/ -- I think it's fixed!

zippy (approved):
Ran it locally and it looks good.  Didn't really dig into the code.

pdaoust:
I'm not entirely sure why it works, but it works :D A bit of recursion in there and it addled my brain.

PR #514, opened 2025-01-20
hotfix: remove mention of non-existent 'next section' in 0.4 upgrade guide; replace with actual code

pdaoust:


jost-s (approved):

alastairong1 (approved):
Looks good. It could be better if we have some list of features or an example with more than one feature flag.

pdaoust:
@alastairong1 for `hdi` and `hdk` there's only one feature flag -- `unstable-functions` -- which enables the functions in the preceding bullet list. If that's not clear I wonder if I should be rewording this section further.

alastairong1:
@pdaoust Are you sure there's only 1 feature flag? In the HDK I needed both `unstable-functions` and `unstable-countersigning` and confirmed it with @ThetaSinner

jost-s:
True, yes, I missed that. It's mentioned under 1 that you need both to compile Holochain, and you also need both under 2.

jost-s:
Depending on which feature you're enabling, that is.  The hdk only has the `unstable-countersigning` feature.

I would actually prefer to not go into the specifics of Rust features here. Instead I'd generically mention that features have to be enabled when compiling Holochain as well as when compiling zomes.

pdaoust:
@alastairong1 okay, looks like `unstable-countersigning` is the only feature flag that got left undocumented, and only in `hdk`. I'll create a new PR.

pdaoust:
@jost any reason you prefer to not mention the specific features? I feel like it'd be super helpful for people with rare but specific needs (not naming any names but @alastairong1 😅 ).

Anyhow #515 goes against your advice but I think it's a helpful addition. Thoughts?

ThetaSinner:
Because people need to know how to use Rust. There are many thousands of pages of documentation that detail the language, Cargo, the standard library. Entire books about idioms and patterns for it. It's just out of scope. We explain how our system works and outside that is documented somewhere else. 

Not least because when these things change in Rust, we don't need our documentation getting invalidated. But for as long as "features" are a concept in Rust, there will be some way to enable them.

PR #515, opened 2025-01-21
hotfix: mention unstable-countersigning flag for hdk in 0.4 upgrade guide

pdaoust:


jost-s (commented):
This change looks good to me. I still think there shouldn't be a code snippet of how to enable features in this guide, just that features need to be enabled in `holochain` as well as `hdk`. Alastair can look it up in [the Cargo Book](https://doc.rust-lang.org/cargo/reference/features.html). =D

jost-s:
I didn't say there doesn't need to be, I said there should not be in my opinion. A busy developer knows how to figure this out.

PR #516, opened 2025-01-21
App structure: DNAs page

### LINKED ISSUE #488: App structure 2: DNA manifest and bundling
* manifest format
    * DNA modifiers vs non-modifiers
    * properties (and overloading at install/clone time -- also app bundling time?)
    * provisioning strategy
    * flags
* using `hc` to bundle a DNA

Broken out from #480.

pdaoust:
All about DNAs. Closes #488. Depends on #512. Don't review until those are merged, cuz otherwise this PR will seem bigger than it really is.


src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+
+If you use the [scaffolding tool](/get-started/3-forum-app-tutorial/), it'll scaffold a working directory for every DNA you scaffold.
+
+You can also use the `hc` command in the [Holonix dev shell](/get-started/#2-installing-holochain-development-environment) to create a bare working directory:
+
+```bash
+hc dna init movies
+```
+
+You'll be prompted to enter a name and [**network seed**](#network-seed). After that it'll create a folder called `movies` that contains a basic `dna.yaml` file with your responses to the prompts.
```
jost-s, line 27: I would focus on the scaffolding command and mention it. The `hc dna` command does work but is of very limited use.

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+  - name: movies
+    hash: null
+    bundled: 'zomes/movies/target/wasm32-unknown-unknown/release/movies.wasm'
+    dependencies:
+    - name: movies_integrity
+```
+
+### DNA manifest structure at a glance
+
+* `name`: A string for humans to read. This might get used in the admin panel of Holochain [conductors](/concepts/2_application_architecture/#conductor) like [Holochain Launcher](https://github.com/holochain/launcher) or [Moss](https://theweave.social/moss/).
```
jost-s, line 59: I wouldn't reference community projects like Moss in the Holochain dev guide.

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+* `integrity`: Contains all the integrity modifiers for the DNA, the things that **change the DNA hash**.
+    * `network_seed`: A string that serves only to change the DNA hash without affecting behavior. It acts like a network-wide passcode. {#network-seed}
+    * `properties`: Arbitrary, application-specific constants. The integrity code can access this, deserialize it, and change its runtime behavior. Think of it as configuration for the DNA.
```
jost-s, line 62:
```suggestion
    * `properties`: Arbitrary, application-specific constants. The coordinator code can read this at runtime Think of it as configuration for the DNA.
```

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+This will create a file in the same folder as the `dna.yaml`, called `<name>.dna`, where `<name>` comes from the `name` field at the top of the manifest.
+
+## Make a coordinator zome depend on an integrity zome
+
+<!-- TODO: write about depending on multiple integrity zomes with `hdk_to_coordinates` -->
+
+In order for a coordinator zome to read and write the entry and link types defined by an integrity zome, you'll need to specify the dependency in a few places.
+
+1. In your coordinator zome's `Cargo.toml` file, specify a dependency on the integrity zome's crate just like you would any Cargo dependency. You can see how to do this in the [Create a coordinator zome section](/build/zomes/#create-a-coordinator-zome) on the Zomes page.
+2. In your DNA manifest file, specify the dependency in the `coordinator` section by referencing the integrity zome's `name` field. You can see an example [above](#specify-a-dna-manifest), where the `movies` zome depends on the `movies_integrity` zome. (Remember that you don't need to do this if there's only one integrity zome.)
```
jost-s, line 93: The dependency applies for one or multiple integrity zomes.
```suggestion
2. In your DNA manifest file, specify the dependency in the `coordinator` section by referencing the integrity zome's `name` field. You can see an example [above](#specify-a-dna-manifest), where the `movies` zome depends on the `movies_integrity` zome.
```

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+
+Then, in your coordinator zome's code, import the integrity zome's entry and link types enums and entry structs/enums as needed:
+
+```rust
+use movies_integrity::{EntryTypes, LinkTypes, Movie, Director};
+```
+
+!!! info Why do I need to specify the dependency twice?
```
jost-s, line 101: This indexing is independent of `dependencies`. `dependencies` are just checked when installing an app bundle to make sure the referenced zomes are available.

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+
+* **Dividing responsibilities.** For instance, a video sharing hApp may have one group of peers who are willing to index video metadata and offer search services and another group of peers who are willing to host and serve videos, along with people who just want to watch them. This DNA could have `search` and `storage` DNAs, along with a main DNA that allows video watchers to look up peers that are offering services and query them.
+* **Creating privileged spaces.** A chat hApp may have both public and private rooms, all [cloned](/resources/glossary/#cloning) from a single `chat_room` DNA. This is a special case, as they all use just one base DNA, but they change just one [integrity modifier](#dna-manifest-structure-at-a-glance) such as the network seed to create new DNAs.
+* **Discarding or archiving data.** Because no data is ever deleted in a cell or the network it belongs to, a lot of old data can accumulate. Creating clones of a single storage-heavy DNA, bounded by topic or time period, allows agents to participate in only the networks that contain the information they need. As agents leave networks, unwanted data naturally disappears.
```
jost-s, line 119: Data from agents that left a network isn't purged. Data from storage-heavy clones can be migrated to another clone and then the old clone can be deleted for example.

src/pages/build/dnas.md
```diff
@@ -0,0 +1,220 @@
...
+    match response {
+        ZomeCallResponse::Ok(data) => data
+            .decode()
+            .map_err(|e| wasm_error!("Couldn't deserialize response into search results: {}", e)),
+        _ => Err(wasm_error!("An unknown error just happened"))
+    }
+}
+```
+
+Note that **bridging between different cells only happens within one agent's hApp instance**, and **remote calls only happens between two agents in one DNA's network**. For two agents, Alice and Bob, Alice can do this:
```
jost-s, line 201:
```suggestion
Note that **bridging between different cells only happens within one agent's hApp instance**, and **remote calls only happen between two agents in one DNA's network**. For two agents, Alice and Bob, Alice can do this:
```

src/pages/build/entries.md
```diff
@@ -113,42 +113,65 @@ let movie = Movie {
...
     // you need to do is wrap your movie in the corresponding enum variant.
     &EntryTypes::Movie(movie),
 )?;
 ```
 
+### Create with relaxed chain top ordering
+
+If your entry doesn't have any dependencies on other data, you can use [relaxed chain top ordering](/build/zome-functions/#relaxed-chain-top-ordering) to prevent possible transaction rollbacks (we'll let that page explain when this could happen and how to design around it).
+
+To use this feature, you'll need to use the more low-level [`create`](https://docs.rs/hdk/latest/hdk/entry/fn.create.html) host function, which requires you to build a more complex input. This example batches updates to director entries, which don't have reference other data including each other, so they're a good candidate for relaxed ordering.
```
jost-s, line 126:
```suggestion
To use this feature, you'll need to use the more low-level [`create`](https://docs.rs/hdk/latest/hdk/entry/fn.create.html) host function, which requires you to build a more complex input. This example batches updates to director entries, which don't have to reference other data including each other, so they're a good candidate for relaxed ordering.
```

src/pages/build/entries.md
jost-s, line 1: Also unexpected to find this addition in this PR.

src/pages/build/zome-functions.md
jost-s, line 1: What's this doing here? I'm beginning to think that you've based this branch off another one that resolve other issues?!

jost-s (changes requested):
Really long PR that includes much more than the issue is about. I suppose this is based off of another branch that addresses the other issues, so I haven't reviewed callbacks and zomes. I did review the DNA part and glossed over the relaxed ordering parts for entries. I haven't run the code examples.

The DNA part looks good to me with a few change requests.

src/pages/build/callbacks-and-lifecycle-hooks.md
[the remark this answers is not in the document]
jost-s: I see now that there's another PR for it. Please mark such dependencies on the PR description.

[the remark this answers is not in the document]
pdaoust: As I was reading your first comment I thought "huh? long PR? It's tiny" and then I realised it's cuz I'm building all these PRs on prior ones so I can keep moving. I'll note deps in the description from now on.

src/pages/build/dnas.md
pdaoust: Hm, looks like we've got an [inaccuracy in the Rustdoc](sudo add-apt-repository ppa:kicad/kicad-8.0-releases) then:
image

src/pages/build/dnas.md
pdaoust: `dna_info` is in the `hdi` crate, so it's accessible to the `validate` and `genesis_self_check` functions. This lets you implement patterns like Progenitor.

src/pages/build/dnas.md
jost-s: I don't see where that would happen. There's just this https://github.com/holochain/holochain/blob/develop/crates/holochain_types/src/dna/dna_manifest.rs#L94

src/pages/build/dnas.md
jost-s: Ah okay, so just "the zome code".

src/pages/build/dnas.md
pdaoust: I'd written a big explanation of my discoveries last week re: how the conductor maps a coordinator's internal entry type zome indices to the order specified in the DNA manifest, but all I found is that it [should happen somewhere in here](https://github.com/holochain/holochain/blob/develop/crates/holochain/src/core/ribosome/host_fn/create.rs#L52-L63) but doesn't. It seems to have gotten lost in one of the refactors.

Now I'm worried that any multi-integrity-zome DNAs are not having internal zome indices mapped to external ones, even when a coordinator zome explicitly depends on only one integrity zome! Going to do some more research.

Here's what I had written. FWIU this is how it used to work, and is supposed to still work:

---

It's rather convoluted. Basically you don't need to specify integrity dependencies for the coordinator zomes at all. All you need are the zome index and entry/link type index and you're good. The problem is that a coordinator can't know the order of integrity zomes in the DNA manifest, so the `depenencies` field sets up a mapping so that when a coordinator tries to call a host function like `create` and specifies an entry type of `(0, 7)` the conductor knows it actually means `(3, 7)` according to its knowledge of the ordering in the DNA manifest.

There's a distressingly magical attribute macro called `hdk_to_coordinates` in the `hdk_derive` crate that does all the zome indexing you, even with multiple integrity deps if you add it to a 'mother' enum like so:

```rust
#[hdk_to_coordinates]
pub enum EntryTypes {
    Foo(integrity_foo::EntryTypes),
    Bar(integrity_bar::EntryTypes),
{

let foo_baz = EntryTypes::Foo(integrity_foo::EntryTypes::Baz(integrity_foo::Baz("baz")));
let foo_baz_index: ScopedEntryDefIndex = foo_baz.into();
debug!("Zome index: {}, entry def index: {}", foo_baz_index.zome_index, foo_baz_index.zome_type);
// Zome index: 0, entry def index: 3
```

src/pages/build/dnas.md
pdaoust: It used to do more than that -- the `create` host invocation used to map local zome indices to its knowledge of global integrity zome ordering (see above comment). I think this is a regression.

src/pages/build/dnas.md
pdaoust: looks like what I'm explaining isn't clear -- I'm trying to saying what you're saying :) I'll update the text and get you to re-review.

src/pages/build/dnas.md
pdaoust: Okay, well, I'm completely confused about how/where local-to-global zome index mapping happens, but I've tested this claim out:

> (Remember that you don't need to do this if there's only one integrity zome.)

And it is correct.

src/pages/build/entries.md
pdaoust: I'm gonna save this fix for #511. (Thank you for being so thorough btw!)

pdaoust: https://github.com/holochain/docs-pages/pull/512#discussion_r1931273097

src/pages/build/dnas.md
jost-s: > And it is correct.

It is correct, but you can also use 2 integrity zomes and not specify any dependencies and it'll work.

The magic has nothing to do with the dependencies specified in the DNA manifest. It's simply that all integrity zomes inside of the DNA are indexed in sequence, and inside each integrity zome the variants of the `hdi_entry_types` decorated enum is indexed as well, and that's what's looked up again when dealing with entry types.

jost-s: A happ dev doesn't need to be concerned about the internal integrity zome lookup. It's got nothing to do with the DNA manifest and there's no way to screw it up.

pdaoust: > but you can also use 2 integrity zomes and not specify any dependencies and it'll work.

It doesn't seem to work; when I leave out the deps I get an error when I try to cast an entry type from either integrity zome to a `ScopedEntryDefIndex` (which `create_entry` does):

```
Wasm runtime error while working with Ribosome: RuntimeError: WasmError { file: "dnas/my_dna/zomes/integrity/second_dep/src/lib.rs", line: 7, error: Guest("Foo does not map to any ZomeIndex and EntryDefIndex that is in scope for this zome.") }
```

> there's no way to screw it up.

I found a way to screw it up.

1. Create a hApp with two integrity zomes, `integrity0` and `integrity1`, and one coordinator zome.
2. In the DNA manifest, specify that the coordinator depends on both integrity zomes.
3. Choose one integrity zome to put as a dep into the coordinator's `Cargo.toml`, doesn't matter which (you can even choose both).
4. No matter which one you pick, any entry type's zome index will be `0`, whereas it should be either `0` or `1` depending on which integrity zome's `EntryTypes` enum you used in your code. (Don't try to use both; that causes a weird compile error and that's a known issue.)
5. Try reversing the order of the dependencies in the DNA manifest. Now all entry types' zome indices will be `1`.

I recognise this is irregular, and doesn't follow the pattern we see devs using in the wild, but we explicitly document that a coordinator may have more than one dependency (and I'd argue that making it a `Vec` is implicitly documenting this fact. So there's either a bug in the documentation/Rust types or the mapping. I'll create an issue for it.

pdaoust: anyhow, it's obvious that this is broken enough that zero- or two-dependency coordinators in a DNA with more than one integrity zome is broken enough to be warned against, so for now I'll just update the documentation.

src/pages/build/dnas.md
pdaoust: Update: turns out `dependencies` is also checked every time you convert an `EntryTypes` value into a `ScopedEntryDefIndex`, and if you have more than one dependency it gets weird ( holochain/holochain#4660 ). I've updated this -- hopefully it's simpler, gives devs a reassurance that Holochain will handle all the mapping, but ONLY IF your coordinator has only one dep.

jost-s: In my opinion this info talks about Holochain implementation details that a happ developer does not need to know about and which could change over time. I'd just write the guide so that coordinator dependencies are always specified and leave out warnings and infos. It can't be true that > 1 dependent integrity zomes is a bug. [mewsfeed uses 2](https://github.com/GeekGene/mewsfeed/blob/main/dnas/mewsfeed/workdir/dna.yaml#L41).

src/pages/build/dnas.md
jost-s: It still says "the integrity code".

src/pages/build/dnas.md
jost-s: Okay, if that's a bug then fine.

jost-s (dismissed):

src/pages/build/dnas.md
pdaoust: so weird; not the first time my edits have just vanished. Thanks for the heads-up.

jost-s (dismissed):

jost-s (approved):

PR #518, opened 2025-01-24
Add a standard CI pass check

ThetaSinner:


.github/workflows/spellcheck.yaml
```diff
@@ -2,6 +2,8 @@ name: 'Check spelling'
 on: # rebuild any PRs and main branch changes
   pull_request:
   push:
+    branches:
+      - main
```
ThetaSinner, line 6: Building twice on every PR and didn't do what the comment said.

.github/workflows/spellcheck.yaml
```diff
@@ -16,4 +18,16 @@ jobs:
         with:
           root: 'src'
           files: '**/*.{json5,njk,md}'
-          incremental_files_only: false
\ No newline at end of file
+          incremental_files_only: false
+
+  ci_pass:
```
ThetaSinner, line 23: Required by Pulumi integration

.github/workflows/spellcheck.yaml
```diff
@@ -16,4 +18,15 @@ jobs:
...
+    if: ${{ always() }}
+    runs-on: "ubuntu-latest"
+    needs:
+      - spellcheck
+    steps:
+      - name: check status
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
```
cdunster, line 32: I never actually know why we add this extra job? Is it just so that we can mark this one as required instead of marking every job as required?

cdunster (approved):

.github/workflows/spellcheck.yaml
ThetaSinner: Yes, and then going one level deeper into the why, it's because we often change what checks happen between branches. That makes having globally required checks is a pain because if you change the required set on `develop-0.4` then builds stop working on `develop-0.3` stops allowing you to merge. That's currently the case on the Holochain repository. 

So this is just about having a common reference point to say "CI checks must pass" without needing to know anything more about what checks are actually being run.

PR #520, opened 2025-02-04
Build guide: validation

### LINKED ISSUE #475: Practical guide for using Membrane Proof
Closed by #520.

Must include:
- code examples
- gotchas

Possible topics to cover:
* Membrane proof 
          * Genesis self-check -- not a 'true' validation function, just a way to guard yourself against copy/paste mistakes and other things that can permanently hose your chance of joining a network
          * (future) Handled specially -- restricts/grants access to a network; validated at handshake time (turns out this is not currently implemented, and there are questions about how to implement it in a way that doesn't carry a performance hit with each new peer connection -- and there may be lots of them in a big heavily sharded DHT)
              * `AgentValidationPkg` is the only action for which an honest person can get warranted, because they try to join the network and publish it before they're able to fetch deps

pdaoust:
Closes #475 .

**This PR depends on #516 and #517 . Don't review until they're merged.**

So while this PR reflects my attempts to simplify the work I'm doing on the Build Guide, it doesn't incorporate the feedback I got today to simplify even more. So it does do some explanatory stuff that may or may not be necessary. Feel free to offer feedback on whether it's too much information, with the caveat that I'll be using that feedback in _future_ PRs and will probably leave this one as it is, because it's already been written and it's probably useful information. (That said, if you feel it gets in the way, please explicitly say that you think it should be removed.)

Covers:

* `validate` callback
* `genesis_self_check` callback
* `must_get_*`
* a few design considerations for validation


pdaoust:
Closes #475 .

src/pages/build/dht-operations.md
jost-s, line 1: I think this background information is interesting but not required in the build guide.

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+
+```rust
+use hdi::prelude::*;
+
+pub fn validate_create_movie(
+    _action: EntryCreationAction,
+    movie: Movie,
+) -> ExternResult<ValidateCallbackResult> {
+    // Note that converting a `&str` to a `Timestamp` requires you to list
+    // `kitsune_p2p_timestamp` as a dependency in your `Cargo.toml`.
```
jost-s, line 71: I think we'll remove or at least move `Timestamp` to a different place, so I suggest to use a std `Instant` in the example instead.

jost-s (commented):
The validation examples look great to me. The pages on DHT operations and validation are out of scope of a build guide IMO.

src/pages/build/genesis-self-check-callback.md
```diff
@@ -0,0 +1,110 @@
...
+title: "Genesis Self-Check Callback"
+---
+
+::: intro
+To enforce access control for a network, a DNA can require a [**membrane proof**](/concepts/3_source_chain/#source-chain-your-own-data-store), which is a piece of data that gets entered by the user and written to their [**source chain**](/concepts/3_source_chain/). The `genesis_self_check` function can guard against user entry error and help prevent them from being banned from the network accidentally.
+:::
+
+## Membrane proof: a per-agent joining code for a network
+
+If your network needs to enforce membership control, your [`validate` callback](/build/validate-callback/) can check the contents of the [`AgentValidationPkg`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/action/enum.Action.html#variant.AgentValidationPkg) record. This [**genesis record**](/resources/glossary/#genesis-records) contains data that's entered by the user on app installation. If validators find an invalid proof, they can ban the unauthorized agent from joining the network.
```
ThetaSinner, line 11: We should be really clear about what "ban" means here. It's not fair on app developers to be unclear about this because it looks like a stronger condition than it is.

We can reject all future ops authored by that agent. This prevents them *writing* data in our DHT. It does NOT prevent *read* access.

This is something we're looking to fix but for now, I'd like the documentation to be clear about this so that people aren't expecting more "security" than they're getting.

src/pages/build/genesis-self-check-callback.md
```diff
@@ -0,0 +1,110 @@
...
+
+A membrane proof can be basic, like per-agent invite codes, or it can be something complex like a [JSON Web Token (JWT)](https://jwt.io/) signed by an agent that has authority to admit members.
```
ThetaSinner, line 13: Are JWTs a recommendation based on something that somebody has done in practice? I'm not saying they couldn't be used but JWTs generally come with an `exp` claim that makes them useless after a period of time. That's no good for Holochain's validation because that can't be dependent on time. So you could skip the recommended claims and use JWTs but then you're really just using digital signatures and you don't need the JWT technology at all.

src/pages/build/genesis-self-check-callback.md
```diff
@@ -0,0 +1,110 @@
...
+
+!!! info Membership control isn't implemented yet
+This feature is not fully implemented. Currently, validators merely record validation failure for a membrane proof and supply it on request. Our plan is to use membrane proof validation outcomes to admit or block communications from peers.
```
ThetaSinner, line 16: Ahh, okay so there is a mention of this here. This is good, but I'd still like to be clearer about about what "ban" means

src/pages/build/genesis-self-check-callback.md
```diff
@@ -0,0 +1,110 @@
...
+This more complex example deserializes a JWT, checks that its [`sub` (subject) field](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2) matches the agent's public key, and checks that its [`iss` (issuer) field](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1) matches an authorized key hard-coded into the DNA's [`properties` block](/build/dnas/#use-dna-properties). {#jwt-membrane-proof}
+
+```rust
+use hdi::prelude::*;
+
+// A type for deserializing a JWT payload from our membrane proof.
+#[derive(Serialize, Deserialize)]
+pub struct MembraneProofJwtPayload {
+    pub iss: AgentPubKeyB64,
+    pub sub: AgentPubKeyB64,
```
ThetaSinner, line 55: This is what I mean here, this is really just down to digital signatures. 

I think you'd do fine here taking an authorized agent pub key from the DNA properties and checking that the membrane proof is a valid signature of the agent key for the agent that is joining.

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+
+Validating an op can have one of three outcomes: {#validation-outcomes}
+
+1. `ValidateCallbackResult::Valid` means that validation succeeded.
+2. `ValidateCallbackResult::Invalid(String)` carries information about the validation failure for debugging.
+3. <code>ValidateCallbackResult::UnresolvedDependencies([UnresolvedDependencies](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/validate/enum.UnresolvedDependencies.html))</code> means that validation couldn't finish because the required dependencies couldn't be fetched from the DHT.
+
+Outcomes 1 and 2 are definitive and the op won't be validated again. Outcome 3, which happens automatically when a call any of the [`must_get_*` functions](https://docs.rs/hdi/latest/hdi/entry/index.html#functions) yields no data, asks Holochain to try running the op through validation later in the hope that it can retrieve the required dependencies. (It'll give up retrying after some time.)
```
ThetaSinner, line 19: It does give up to save resources but if you restart Holochain it will resume trying. Abandon is not implemented as a validation outcome.

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+pub fn validate(_: Op) -> ExternResult<ValidateCallbackResult> {
+    Ok(ValidateCallbackResult::Invalid("I reject everything".into()))
+}
+```
+
+There is something worth noting here though. We return an `Ok` even though validation failed. That's because **`Err` should be reserved for true failures** such as the ones returned by host functions.
+
+## Create boilerplate code with the scaffolding tool
+
+Even after flattening, there are a lot of variants of an op, which makes for a lot of match arms in your `validate` callback. And a lot of the arms do similar but slightly different things. It's a great idea to let the scaffolding tool generate the callback for you. We won't paste a sample here --- you can scaffold an integrity zome and a few entry and link types and see what it looks like.
```
ThetaSinner, line 47: Was flattening introduced before here?

src/pages/build/validate-callback.md
ThetaSinner: This has now moved to `holochain_timestamp`, I think it's probably okay to reference that now?

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+```rust
+use hdi::prelude::*;
+
+pub fn validate_create_movie(
+    _action: EntryCreationAction,
+    movie: Movie,
+) -> ExternResult<ValidateCallbackResult> {
+    // Note that converting a `&str` to a `Timestamp` requires you to list
+    // `kitsune_p2p_timestamp` as a dependency in your `Cargo.toml`.
+    if movie.release_date < "1888-10-14".try_into().unwrap() {
```
ThetaSinner, line 72: Rather than having to parse a date for every movie that gets created, I'd figure out the UNIX timestamp for that data and put it in a constant. Then it's just `Timestamp::from_micros`

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+
+pub fn validate_create_movie(
+    _action: EntryCreationAction,
+    movie: Movie,
+) -> ExternResult<ValidateCallbackResult> {
+    // Just call the function, and Holochain will handle the
+    // `UnresolvedDependencies` outcome for you.
+    let director_entry =  must_get_entry(movie.director_hash)?;
+    // Try to turn it into an entry of the right type.
+    let _director = crate::Director::try_from(director_entry)?;
```
ThetaSinner, line 92: Has this been tested? I always struggle to keep how validation works in my head but I think this is expressing the wrong thing. This will be treated as a code error and *not* as failed validation. I think you need to explicitly handle this error and return an app validation rejected outcome if this doesn't point to a Director type. 

You could check what happens here if you write a coordinator function that tries to reference another movie in the `director_hash` for example

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+    Ok(ValidateCallbackResult::Valid)
+}
+```
+
+#### `validate_update_<entry>`
+
+This callback gets the original entry and its creation action along with the update. You can use it to compare differences between the two entries, or you can use it to enforce write permissions, such as only allowing the original author to update an entry:
```
ThetaSinner, line 99: Definitely worth noting that people should consider what this logic means. For a profile picture, it makes a lot of sense to restrict updates to the original author. For movies, I could claim the name of the movie in the app because I heard about it being filmed before it was released. Just a quick note here about the app validation rules being part of the business logic to an extent.

OR

In this example, you'd probably want to check that the create rules aren't violated by the update, and allow other people to correct wrong movie details as long as they follow the rules.

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,226 @@
...
+        Some(invite_code) => {
+            decode_invite_code(invite_code)?;
+            Ok(ValidateCallbackResult::Valid)
+        }
+        None => Ok(ValidateCallbackResult::Invalid("Please supply an invite code.".into()))
+    }
+}
+```
+
+<!-- TODO: Write about inductive validation when it's not broken https://github.com/holochain/holochain/issues/4669 -->
```
ThetaSinner, line 208: I don't follow. Validation is inductive when `must_get_valid_x` is used is it not?

Warrants are unstable, packed with bugs and disabled by default. I wouldn't let them stop you writing about the logic of depending on existing valid data.

src/pages/build/validation.md
```diff
@@ -0,0 +1,79 @@
...
+
+A DNA uses validation logic in two ways:
+
+1. By an author of data, to protect them from publishing invalid data, and
+2. By an agent that's received data to store and serve, to equip them to detect invalid data and take action against the author.
+
+Because every peer has the DNA's validation logic on their own machine and is expected to check the data they author before they publish it, it can be assumed that invalid data is (almost) always an intentionally malicious act.
+
+!!! info
+Currently Holochain can inform agents about invalid data when asked. In the future it'll also take automatic defensive action by putting a malicious author into an agent's network block list when they see evidence of invalid data.
```
ThetaSinner, line 28: Is that not already happening? It's supposed to be the case that we block people. The missing piece was warrants so that agents would be properly blocked in a sharded network where not everyone would see the invalid ops to discover that the agent needed blocking.

ThetaSinner (commented):
There are two things missing for me from this. One is that although Holochain prevents you from calling host functions that prevent you from getting the current time or other non-deterministic things, nothing stops you doing non-deterministic things in your code. It's up to you to ensure that you follow the intentions that Holochain has set out. I think that might be worth mentioning somewhere in the general validation page.

Then there's also testing. Testing is hugely important. It's far easier than it seems to write validation rules that reject valid data or fail to progress through data. You need to test that you can take all the actions that you expect but it's also important to write Tryorama tests that take those some actions on a network with at least two agents and check that the DHT syncs between the agents. This is from personal experience of writing validation rules that pass for the author but not for other peers.

src/pages/build/validate-callback.md
pdaoust: @ThetaSinner has it moved there for the 0.4 series, or just 0.5?

src/pages/build/dht-operations.md
```diff
@@ -0,0 +1,56 @@
...
+
+An [action](/build/working-with-data/#entries-actions-and-records-primary-data) on an agent's [source chain](/concepts/3_source_chain/) yields multiple DHT operations, each of which goes to an [**authority**](/resources/glossary/#validation-authority) for that operation's [**basis address**](/resources/glossary/#basis-address) (a DHT address that the authority is responsible for).
+
+Here are all the DHT operations produced for all the actions:
+
+* All actions
+    * [`RegisterAgentActivity`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/op/enum.Op.html#variant.RegisterAgentActivity)
+        * Basis address: author's public key
+        * Contents: action (and optionally entry, if applicable)
+        * Effect: Store the action. Because one agent's actions will all go to the same set of authorities, their job is to check for [source chain forks](/resources/glossary/#fork-source-chain), which is a system-level rule that you don't have to implement.
```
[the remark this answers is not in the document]
pdaoust: Re: too much detail, I agree, and it's this page I was thinking of when I said

> So it does do some explanatory stuff that may or may not be necessary. [...] it's already been written and it's probably useful information.

It used to be on the validation page but was TMI, so I thought I'd hide it on its own page where it won't hurt anyone. Still seems useful for people wanting to get advanced, and I'd probably end up writing it a year down the road anyway. Do you think it should be removed entirely?

pdaoust:
@ThetaSinner 

> nothing stops you doing non-deterministic things in your code

I wracked my brain to figure out how that might be possible -- any source of entropy (like `hdk::random::random_bytes` or `hdk::time::sys_time`) or anything that varies by agent -- and I can't think of anything. Can you? Are you thinking of third-party libraries?

src/pages/build/callbacks-and-lifecycle-hooks.md
```diff
@@ -21,44 +21,16 @@ The `validate` callback is called at two times:
...
-
-There's one challenge with this: validation requires access to the network (a membrane proof may reference DHT data that contains a list of public keys authorized to grant access to newcomers, for instance), but the membrane proof is written at [**genesis**](/resources/glossary/#genesis-records) time when the cell doesn't have network access yet. So Holochain can't do the usual self-validation before publishing it.
-
-This means someone could accidentally type or paste a malformed membrane proof and be rejected from the network. To guard against this, you can define a `genesis_self_check` function that runs at genesis time and checks the content of the membrane proof before it's written.
+There's a moment in a cell's life, after it's been instantiated but before it's connected to its network, where it's published data that it can't fully validate. This data is their [**membrane proof**](/concepts/3_source_chain/#source-chain-your-own-data-store), an optional chunk of data that can serve as a joining code for the network. You can write validation rules for it just like any other record, allowing existing agents to reject newcomers with invalid credentials.
```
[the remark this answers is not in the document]
pdaoust: oh yeah, it is redundant. I like your rewrite more though; gonna delete the second instance, if you don't mind :)

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
pdaoust: (IOW, this _is_ the advanced guide and I warn people about that fact before they reach this page)

src/pages/build/dht-operations.md
```diff
@@ -0,0 +1,56 @@
...
+* [`DeleteLink`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/action/enum.Action.html#variant.DeleteLink)
+    * [`RegisterDeleteLink`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/op/enum.Op.html#variant.RegisterCreateLink)
+        * Basis addresses: old link's [base address](/build/links-paths-and-anchors/#define-a-link-type) and action hash
+        * Contents: action
+        * Effect: Mark a link as deleted, without removing the actual data.
+
+### Choosing who should validate what
+
+In practice, it's usually okay to have all groups of authorities validate all the data in an action. However, if your validation logic is highly complex and computationally costly, it can sometimes be useful to choose different validation tasks for the different DHT operations. For instance, a `RegisterAgentActivity` authority may not need to validate entry data; it could just focus on the action in the context of the entire source chain. Or a `StoreEntry` authority could focus on entry contents and leave checking of create or update privileges to `StoreRecord` authorities.
```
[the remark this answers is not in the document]
pdaoust: hmmm, that's the thing I'm trying to convey, or at least allude to. I'll edit it.

src/pages/build/genesis-self-check-callback.md
```diff
@@ -0,0 +1,110 @@
+---
+title: "Genesis Self-Check Callback"
+---
+
+::: intro
+To enforce access control for a network, a DNA can require a [**membrane proof**](/concepts/3_source_chain/#source-chain-your-own-data-store), which is a piece of data that gets entered by the user and written to their [**source chain**](/concepts/3_source_chain/). The `genesis_self_check` function can guard against user entry error and help prevent them from being banned from the network accidentally.
+:::
+
+## Membrane proof: a per-agent joining code for a network
```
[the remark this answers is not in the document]
pdaoust: True; as you say, that would be implementation-specific. I don't quite understand the concern; does it sound like 'per-agent' is suggesting that it _has to_ be different for each agent?

I've already mentioned that a network seed is like a network-wide passcode, so I want some way of distinguishing the membrane proof here. Happy to entertain alternative wording tho!

src/pages/build/genesis-self-check-callback.md
pdaoust: heh, fair point. Guess I'm nodding to the future of what this thing is for. Get that it's only an agent-centric ban too, i.e., adding the agent to one's own blocklist. That's not implemented (or rather, it's been turned off with a feature flag), right?

src/pages/build/genesis-self-check-callback.md
pdaoust: Nah, just looking for something that devs would be familiar with. Happy to create a simpler structure and skip the base64 encoding junk.

src/pages/build/validate-callback.md
pdaoust: oh! didn't know about either of those things.

src/pages/build/validate-callback.md
pdaoust: whoops, casualty of a reorganisation

src/pages/build/validate-callback.md
pdaoust: hm, yeah, that should be a validation failure, not an error. Pretty sure it's gonna do what you think it's gonna do.

src/pages/build/validate-callback.md
pdaoust: Okay sure, so it's not something you're worried about overpromising on? For me the concern is not whether warrants are buggy; it's my concern that, even once warrant bugs are fixed, `must_get_valid_record` is inherently non-deterministic -- we're choosing to relinquish our right to validate the dep and trust the word of a potentially Byzantine-faulty peer.

src/pages/build/validation.md
pdaoust: I dunno; I thought blocking was hidden behind a feature flag. Is that only for coordinator-level blocking?

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
pdaoust: Also: this is relevant even before the network its the sharding threshold, isn't it? If everyone is an authority for everything and the same logic is shared between, say, the five ops produced by an `Update`, then everyone will run the same validation logic five times.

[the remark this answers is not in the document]
pdaoust: No, you can't decide what kind of authority you are, but the hApp dev can decide what logic you're going to run when you do assume authority X. If the same validation logic is duplicated five times for each op that an `Update` produces, the network runs the same logic 5 times × the sharding threshold. This matters even before sharding.

The scaffolded validation code employs this somewhat -- I forget in what ways, but it does divide the responsibility up a bit so work isn't duplicated so much. I'll rewrite this to make it clearer that authorities don't get to choose their ops, but hApp devs do get to choose how the different ops get validated.

src/pages/build/dht-operations.md
```diff
@@ -0,0 +1,56 @@
...
+!!!
+
+An [action](/build/working-with-data/#entries-actions-and-records-primary-data) on an agent's [source chain](/concepts/3_source_chain/) yields multiple DHT operations, each of which goes to an [**authority**](/resources/glossary/#validation-authority) for that operation's [**basis address**](/resources/glossary/#basis-address) (a DHT address that the authority is responsible for).
+
+Here are all the DHT operations produced for all the actions:
+
+* All actions
+    * [`RegisterAgentActivity`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/op/enum.Op.html#variant.RegisterAgentActivity)
+        * Basis address: author's public key
+        * Contents: action (and optionally entry, if applicable)
```
pdaoust, line 21: @ThetaSinner @jost-s is this actually true? I did a quick search and it looks like `EntryDef::cache_at_agent_activity` isn't honoured anywhere in core.

src/pages/build/validate-callback.md
pdaoust: The scaffolded `validate` callback calls `validate_create_movie` for a `StoreEntry` op and both `validate_create_movie` and `validate_update_movie` for a `StoreRecord` op, so I think it'd be redundant to have the example check that the create rules aren't violated. I'd like to show something that's unique to updates... you're right that this is an anti-pattern if you're worried about squatters though. Maybe I'll show an example using `MovieLoan` from the 'Identifiers' build guide instead.

src/pages/build/validate-callback.md
ThetaSinner: Good point, only on 0.5

ThetaSinner: Although for this purpose, it doesn't really matter, you'd just have to compare the raw times with `.as_micros` because you wouldn't be able to compare a K1 timestamp with a Holochain timestamp. Maybe best left alone for now

src/pages/build/validation.md
ThetaSinner: It probably is hidden in the HDK because it's not properly tested but it's all one feature and Holochain should be free to use it.

Looking at the workflows though, I don't see it happening. It logs that a peer is going to be blocked and doesn't do it... I'm going to try and tell myself I'm looking in the wrong place so I'm not grumpy about this all weekend but that looks a lot like another gap in Holochain.

ThetaSinner:
> I wracked my brain to figure out how that might be possible -- any source of entropy (like hdk::random::random_bytes or hdk::time::sys_time) or anything that varies by agent -- and I can't think of anything. Can you? Are you thinking of third-party libraries?

```

**Comment by ThetaSinner** [comment]
`

**Comment by ThetaSinner** [comment]
> Testing is a big subject that we probably ought to add to the build guide. Would a simple mention to "test test test -- we'll write more about this later but for now, look at the stub tests that the scaffold tool generated for you" be enough, do you think?

I'd like to just say something like I did. We suggest testing all your validation rules with a single agent, then also testing them with at least two agents and making sure the DHT syncs. For sure refer them to the generated tryorama tests for how to actually do that. I'm pretty sure it comes with multi-agent tests.

**Comment by ThetaSinner** [comment]
Sorry for the close/open - keyboard shortcuts while trying to enter code

File: src/pages/build/validate-callback.md (hunk 1 above)
Comment (pdaoust) [reply src/pages/build/validate-callback.md:71]: Oh wait -- there's a Holochain timestamp type? I thought there was only a K1 timestamp that was re-exported by `holochain_integrity_types`

File: src/pages/build/validate-callback.md (hunk 8 above)
Comment (pdaoust) [reply src/pages/build/validate-callback.md:72]: So far in the build guide we've been turning strings into kitsune timestamps for legibility's sake; if we decide to change that, I'd like to do it in a separate PR.

File: src/pages/build/genesis-self-check-callback.md (hunk 3 above)
Comment (pdaoust) [reply src/pages/build/genesis-self-check-callback.md]: (although, if it's coming in as a membrane proof, there still needs to be a base64-decoding step -- it's a lot simpler than decoding the JWT, though, which involves two base64-decoding step plus a JSON deserialising step)

**Comment by pdaoust** [comment]
@ThetaSinner 

> ```
> some code
> ```
>
>This builds fine inside an integrity zome but is a really bad idea

Oh my goodness. That is an eldritch horror. Okay, I'll add something to the docs.

File: src/pages/build/validate-callback.md (hunk 1 above)
Comment (ThetaSinner) [reply src/pages/build/validate-callback.md:71]: Yes but K1 is being deleted, so the type has been moved. We should never have been using a Kitsune type in hApp code.

File: src/pages/build/dht-operations.md (hunk 15 above)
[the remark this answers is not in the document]
Comment (ThetaSinner) [reply src/pages/build/dht-operations.md]: Okay, yes but I still disagree with what's written here. As far as I know, we have 3 authority types which are agent, record and entry. You do naturally validate different things at those authorities. For example, if you were going to validate that posts have a unique title by author then you should do that at the agent activity authority where you are holding that agents previous data already. Doing so at the record authority would require fetching and caching agent activity for agents that you happen to be a record authority for which would change over time because there's no association between agent and record locations. Holochain has no provision for resolving a mismatch like this currently where different types of authority disagree on the validity of data. But at least with logic like this, there's a way we could go about that. Trying to puzzle out what it means for part of a network to say a post is valid with or without a title just hurts my head. I can work with "This record is valid, but not published by that author". 

Maybe we should be getting feedback from people like mattyg and guillem here but I'm inclined to suggest that validation of the *content* should always be done. So entry authorities do the least work. Then record and agent authorities build on that validation to check what is appropriate for them. Then we can always say that data that is being served follows the basic rules of the game.

I'd also say that we shouldn't worry about duplication of effort. That is part of the design of Holochain and apps are supposed to scale to a large number of users. If your app design is targeting a small number of users in a DHT then validation is less important IMO. For example, a 1-1 messaging app might just check message length with a character limit. A social media app intended for thousands of people should probably always ensure that a post has the right structure, whoever is serving the data because reasonably quickly, you will only be an authority for a smaller part of the DHT.

File: src/pages/build/validate-callback.md (hunk 8 above)
Comment (ThetaSinner) [reply src/pages/build/validate-callback.md:72]: Something like 

```rust
/// The timestamp for the date 1888-10-14, at 00:00
const EARLIEST_MOVIE_TIMESTAMP: Timestamp = Timestamp(1209483914);
```

Seems perfectly legible to me. I wouldn't reject it on a PR into Holochain.

src/pages/build/validate-callback.md
ThetaSinner: We never trust a single peer on a warrant, that's not the right way to think about it. If somebody who is an authority for some data sees a warrant then they perform the same validation that the author of the warrant did. If that validation also results in them wanting to issue a warrant then they are supposed to block the author of the original data. So what should happen is that all authorities for an invalid record end up with a warrant and block the author. This means that anybody who isn't an authority for that data can't get the record from any authority. There should be no case where we just trust a single warrant issuer. The point of treating it as unresolved dependencies is to see if we can retrieve that dependency from somebody else if one agent falsly claims a warrant. Whether it's correctly implemented is another question but we spent a lot of time puzzling out these cases.

I want to forget completely about warrants here. They're disabled until further notice. The logic of "an update is only valid only against a valid create" is entirely implementable today on Holochain by using `must_get_valid_record`.  I don't see why you'd need to re-check the create when validating an update. Using `must_get_valid_record` is the right way to do that.

src/pages/build/dht-operations.md
ThetaSinner: There shouldn't be an entry with a register agent activity

ThetaSinner: That feature isn't implemented AFAIK

src/pages/build/callbacks-and-lifecycle-hooks.md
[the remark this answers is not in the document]
ThetaSinner: This looks like it hasn't changed?

pdaoust:
@ThetaSinner this commit https://github.com/holochain/docs-pages/pull/520/commits/b19615cbff0f912f56b406bc0f527c88c2af3151 warns about non-deterministic stuff.

src/pages/build/genesis-self-check-callback.md
pdaoust: * https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-c14599996cdb7d4e9167c9e2177448afca23015bcda3b5358cf1cf23560425b2R9
* https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-c14599996cdb7d4e9167c9e2177448afca23015bcda3b5358cf1cf23560425b2R16
* https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-c14599996cdb7d4e9167c9e2177448afca23015bcda3b5358cf1cf23560425b2R21

src/pages/build/genesis-self-check-callback.md
pdaoust: simplified in https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-c14599996cdb7d4e9167c9e2177448afca23015bcda3b5358cf1cf23560425b2R46-L92

src/pages/build/genesis-self-check-callback.md
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-c14599996cdb7d4e9167c9e2177448afca23015bcda3b5358cf1cf23560425b2R46-L92

src/pages/build/validate-callback.md
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-6292d0d84462fb21a3681bc06fadd4f6cc525f5bc93015e51e25adf0e0c2b622R19

src/pages/build/validate-callback.md
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/15dae3589a940e21125190542cdc9f1040fa2ee7#diff-6292d0d84462fb21a3681bc06fadd4f6cc525f5bc93015e51e25adf0e0c2b622R47-R49 , also includes some removal of redundant prose

src/pages/build/validate-callback.md
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/afe9fe11bf816defd19aaeb941a4868798646a61

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/4de1af4957f521992d9e8a7b62eab50c7a5f9835 is a partial attempt to respond to the concern, as well as an attempt to untangle some text that seemed unclear.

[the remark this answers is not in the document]
ddd-mtl: I think its just the title `Choosing who should validate what` that is confusing as in the zome code you can't choose agents in anyway. I'd say the phrasing would be more about load balancing or something of sorts.
I believe any scaffolding related things should be kept in the scaffolding docs.

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Stub functions for entries
+
+These functions can be found in the file `dnas/<dna>/zomes/integrity/<zome>/src/<entry>.rs`.
+
+#### `validate_create_<entry>`
+
+In this function you can write rules for the contents of your entries and their dependencies. You can also write rules for the actions that write them, which means you can create rules for write privileges. (Note that it's called for both kinds of [`EntryCreationAction`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/op/enum.EntryCreationAction.html) --- `Create` and `Update`.)
+
+This example checks that a movie is within [sensible bounds](https://en.wikipedia.org/wiki/Roundhay_Garden_Scene).
```
mattyg, line 61:
```suggestion
This example checks that a movie's release date is within [sensible bounds](https://en.wikipedia.org/wiki/Roundhay_Garden_Scene).
```

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,239 @@
...
+```rust
+use hdi::prelude::*;
+
+pub fn validate_create_movie(
+    _action: EntryCreationAction,
+    movie: Movie,
+) -> ExternResult<ValidateCallbackResult> {
+    // Just call the function, and Holochain will handle the
+    // `UnresolvedDependencies` outcome for you.
+    let director_entry =  must_get_entry(movie.director_hash)?;
```
mattyg, line 91:
```suggestion
    let director_entry = must_get_entry(movie.director_hash)?;
```

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,239 @@
...
+}
+```
+
+You can find other stub functions in that file for links that point to the most recent update (if you [chose that option](/get-started/3-forum-app-tutorial/#scaffold-most-recent-update-link) when you scaffolded the entry type), links that manage collections of entries, and backlinks from entries to entries that depend on them.
+
+### `validate_agent_joining`
+
+Use this function to validate the [**membrane proof**](/build/genesis-self-check-callback/#membrane-proof-a-joining-code-for-a-network). Note that this is different from `genesis_self_check`, in that it's called from the `validate` function so it can access DHT data.
+
+This example implements a simple invite code for a network that people can invite their friends to join. All that's required is the presence of an 'invite' action on the DHT, whose hash becomes the invite code. (It's not a very secure pattern; please don't duplicate this in high-security hApps.) As a bonus, it shows a good pattern where the code for basic pre-validation is shared with `genesis_self_check`.
```
mattyg, line 161:
```suggestion
This example implements a simple invite code for a network that people can invite their friends to join. All that's required is the presence of an 'invite' action on the DHT, whose hash becomes the invite code. (It's not a very secure pattern; please don't duplicate this in high-security hApps.) Some of the business logic is shared with `genesis_self_check`.
```

mattyg, line 161:
```suggestion
This example implements a simple invite code for a network that people can invite their friends to join. All that's required is the presence of an 'invite' action on the DHT, whose hash becomes the invite code. As a bonus, it shows a good pattern where the code for basic pre-validation is shared with `genesis_self_check`.
```

Why not just validate that the invitation includes the recipient agent? Would it be secure then, or is there something else missing?

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,239 @@
...
+
+```rust
+use hdi::prelude::*;
+use base64::prelude::*;
+
+pub fn validate_agent_joining(
```
mattyg, line 167: It's definitely a bit confusing to me that this uses functions `validate_invite_code_format` and `decode_invite_code` that haven't been defined anywhere.


Also it doesn't look like any of the business logic has been shared with `genesis_self_check`?

src/pages/build/validate-callback.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+<!-- TODO: Write about inductive validation when it's not broken https://github.com/holochain/holochain/issues/4669 -->
+
+## Reference
+
+* [`holochain_integrity_types::op::Op`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/op/enum.Op.html)
+* [`holochain_integrity_types::action::AgentValidationPkg`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/action/enum.Action.html#variant.AgentValidationPkg)
+* [`holochain_integrity_types::genesis::GenesisSelfCheckData`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/genesis/type.GenesisSelfCheckData.html)
+* [`holochain_integrity_types::validate::ValidateCallbackResult`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/genesis/type.GenesisSelfCheckData.html)
```
mattyg, line 228:
```suggestion
* [`holochain_integrity_types::validate::ValidateCallbackResult`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/validate/enum.ValidateCallbackResult.html)
```

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+    * [`genesis_self_check` Callback](/build/genesis-self-check-callback/) --- writing a function to control access to a network
+    * [`validate` Callback](/build/validate-callback/) --- basic callback, examples using stub functions
+    * [DHT operations](/build/dht-operations/) --- advanced details on the underlying data structure used in DHT replication and validation
+:::
+
+::: intro
+Validation gives shape to your [DNA](/build/dnas/)'s data model. It defines the 'rules of the game' for a network --- who can create, modify, or delete data, and what that data should and shouldn't look like. It's also the basis for Holochain's peer-auditing security model.
+:::
+
+You implement your validation logic in your application's [integrity zomes](/build/zomes/#integrity). While [entry and link types](/build/working-with-data/) enumerate the kinds of data the integrity zome defines, the data is just bytes until your validation logic gives it some shape and purpose.
```
mattyg, line 18:
```suggestion
You implement your validation logic in your application's [integrity zomes](/build/zomes/#integrity).
```
IMO this is more confusing than clarifying.

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+
+A DNA uses validation logic in two ways:
+
+1. By an author of data, to protect them from publishing invalid data, and
+2. By an agent that's received data to store and serve, to equip them to detect invalid data and take action against the author.
+
+Because every peer has the DNA's validation logic on their own machine and is expected to check the data they author before they publish it, it can be assumed that invalid data is (almost) always an intentionally malicious act.
```
mattyg, line 25:
```suggestion
Because every peer has the DNA's validation logic on their own machine and is expected to check the data they author before they publish it, it can be assumed that invalid data is always an intentionally malicious act.
```

Is there a case where a user can write invalid data unintentionally?

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+
+!!! info
+Currently Holochain can inform agents about invalid data when asked. In the future it'll also take automatic defensive action by putting a malicious author into an agent's network block list when they see evidence of invalid data.
+!!!
+
+There are two callbacks that implement validation logic:
+
+* `validate` is the core of the zome's validation logic. It receives a [**DHT operation**](/build/dht-operations/) and returns a success/failure/indeterminate result.
+* [`genesis_self_check`](/build/genesis-self-check-callback/) 'pre-validates' an agent's [**membrane proof**](/concepts/3_source_chain/#source-chain-your-own-data-store) before trying to connect to peers in the network.
```
mattyg, line 34:
```suggestion
* [`genesis_self_check`](/build/genesis-self-check-callback/) 'pre-validates' your own agent's [**membrane proof**](/concepts/3_source_chain/#source-chain-your-own-data-store) before trying to connect to peers in the network.
```

Is that clearer?

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+* Validation rules should **always yield the same true/false outcome** for a given operation regardless of who is validating them and when. Don't use any source of non-determinism, such as instantiating and comparing two `std::time::Instant`s. In fact Holochain prevents your validation callbacks from calling any non-deterministic host functions. Read more about the [available host functions](#available-host-functions).
+* Data may have dependencies that affect validation outcomes, but those dependencies must be [addressable](/build/identifiers/), they must be retrievable from the same DHT, and their addresses must be constructable from data in the operation being validated. (Note that because an action references the action preceding it on an agent's source chain, this is a dependency you don't need to build into your data.) If dependencies can't be retrieved at validation time, the `validate` callback terminates early with an [indeterminate result](/build/validate-callback/#validation-outcomes), which will cause Holochain to try again later.
```
mattyg, line 43:
```suggestion
* Data may have dependencies that affect validation outcomes, but those dependencies must be [addressable](/build/identifiers/), they must be retrievable from the same DHT, and their addresses must be known. (Note that because an action references the action preceding it on an agent's source chain, this is a dependency you don't need to build into your data.) If dependencies can't be retrieved at validation time, the `validate` callback terminates early with an [indeterminate result](/build/validate-callback/#validation-outcomes), which will cause Holochain to try again later.
```

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+* When multiple actions are written in a same [atomic transaction](/build/zome-functions/#atomic-transactional-commits), actions' ops can only have dependencies on prior actions committed in the transaction, not later actions.
```
mattyg, line 44:
```suggestion
* Despite that multiple actions can be written within an [atomic transaction](/build/zome-functions/#atomic-transactional-commits), they are *not* validated together as an atomic transaction. An action can only have dependencies on prior actions, not subsequent actions.
```

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
+* You don't need to validate your data manually before committing --- Holochain [validates it after the zome function that writes it is finished](/build/zome-functions/#validate-dht-operations), and returns any validation failure to the caller.
+* **Test, test, test.** Validation is the gate that accepts or rejects all DHT data, so make sure you write thorough test coverage for your validation functions. If the data being validated has no dependencies on DHT data or DNA/zome info, we recommend writing Rust unit tests for the [validation function stubs](/build/validate-callback/#create-boilerplate-code-with-the-scaffolding-tool) that the scaffolding tool generates. We also recommend testing your validation code by writing single- and multiple-agent [Tryorama](https://github.com/holochain/tryorama/) test scenarios for zome functions that write data (we'll write about Tryorama soon; in the meantime, you can check the Tryorama GitHub readme and the scaffolded tests in a project's `tests/src/` folder for guidance).<!-- TODO: link to Tryorama page -->
```
mattyg, line 46:
```suggestion
* **Test, test, test.** Validation is the gate that accepts or rejects all DHT data, so make sure you write thorough test coverage for your validation functions. We recommend testing your validation code by writing [Tryorama](https://github.com/holochain/tryorama/) test scenarios for zome functions that write data (we'll write about Tryorama soon; in the meantime, you can check the Tryorama GitHub readme and the scaffolded tests in a project's `tests/src/` folder for guidance).<!-- TODO: link to Tryorama page -->
```

src/pages/resources/glossary.md
```diff
@@ -429,13 +429,21 @@ To create alternate versions of one's history in an app by basing two [source ch
 
 In Holochain terms, synonymous with [graphical user interface](#graphical-user-interface-gui) or, more generally, [client](#client).
 
+#### Genesis
+
+The time period in the lifecycle of a [cell](#cell) in which [genesis records](#genesis-records) are being written. This happens before the cell has access to its [DNA](#dna)'s network and [DHT](#distributed-hash-table-dht) data. A [genesis self-check callback](#genesis-self-check-callback) can be written to pre-validate the [membrane proof](#membrane-proof) during this period.
+
 #### Genesis records
 
 The four records at the beginning of an [agent's](#agent) [source chain](#source-chain), consisting of:
```
mattyg, line 438:
```suggestion
The three records at the beginning of an [agent's](#agent) [source chain](#source-chain), consisting of:
```
Or am I missing something?

src/pages/build/validation.md
```diff
@@ -0,0 +1,81 @@
...
```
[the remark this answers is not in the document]
ThetaSinner: > actions' ops can only have dependencies on prior actions committed in the transaction, not later actions.

Technically this is a *should not* rather than a *cannot* but probably best to document it that way :D Structural references that sys validation will check must be in the right direction. There's nothing stopping you mixing things up in app validation but it will slow down validation and if you create a circular dependency then it will stall validation.

> we recommend writing Rust unit tests for the

I personally wouldn't do this. Constructing the Holochain types as test data is not straight forward. You actually need to know quite a lot about how to construct them for this to work and even then, you have to know the inputs to validation well enough to construct good test cases. It can be done, I wouldn't bother.

src/pages/build/validation.md
ThetaSinner, line 46:
```suggestion
* **Test, test, test.** Validation is the gate that accepts or rejects all DHT data, so make sure you write thorough test coverage for your validation functions. If the data being validated has no dependencies on DHT data or DNA/zome info, we recommend writing Rust unit tests for the [validation function stubs](/build/validate-callback/#create-boilerplate-code-with-the-scaffolding-tool) that the scaffolding tool generates. We also recommend testing your validation code by writing single- and multiple-agent [Tryorama](https://github.com/holochain/tryorama/) test scenarios for zome functions that write data. This lets you check that your validation rules pass both when authoring data and checking data authored by other agents. (we'll write about Tryorama soon; in the meantime, you can check the Tryorama GitHub readme and the scaffolded tests in a project's `tests/src/` folder for guidance).<!-- TODO: link to Tryorama page -->
```

ThetaSinner: Keen to include the why there. We can skip going into detail about how

src/pages/build/validate-callback.md
ThetaSinner: There's nothing much we can do about that. I think it's acceptable for data that is otherwise valid but happens to depend on invalid data to be marked valid. Other agents might not be able to validate it to check your work but that doesn't seem too serious. It's ultimately business logic for the application to decide how/whether to prune out data that ended up relying on invalid data I think. Whether that's practical is another issue. This is very much on the fringes and outside what most apps need to worry too much about I think.

src/pages/build/validate-callback.md
ThetaSinner: The `let = ... else` syntax will only do something useful if the assignment is doing pattern matching. So doesn't that need to be `let Ok(_) = .... else`? I'd expect the compiler to say the else is unreachable but maybe there's something I don't know there.

Linking diffs is helping but when there's that much changed in one commit it's less helpful. I'm assuming that commit went up before we discussed posting diff links but just FYI :)

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
ThetaSinner: I like the re-write better. It reads like it's encouraging the user to think about the effects of validation and giving them a way in to thinking about that.

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
ThetaSinner: This section I consider advanced and also not clearly/correctly implemented in Holochain, the ones above I'd be happy to read about. If it's already been suggested that this becomes its own page then fine, let's keep it that way. 

I am inclined to say that this page is outside the scope of the original issue. I think the `intro` section would be wanted to introduce what validation is operating on but otherwise, authorities are a complex topic.

src/pages/build/dht-operations.md
jost-s: Fine by me

jost-s (dismissed):

src/pages/build/validate-callback.md
pdaoust: already caught that and fixed in https://github.com/holochain/docs-pages/pull/520/commits/11aa2217a77ee4da99a446763d4299b5bade0915#diff-6292d0d84462fb21a3681bc06fadd4f6cc525f5bc93015e51e25adf0e0c2b622L92-R95 .

I don't think I caught the request to link to individual commits rather than diffs, but I've been independently starting to move towards tiny commits that only change one thing, cuz it seems like it fits better with this sort of pattern.

src/pages/build/validation.md
pdaoust: https://github.com/holochain/docs-pages/pull/520/commits/faf557f1eadb372666a4ab4bfba9fb56ae7320d8

src/pages/build/validate-callback.md
pdaoust: Both fixed in https://github.com/holochain/docs-pages/pull/520/commits/a9292c0d1954b6cf3850c5367aab62ba5bce09d2 . The absent functions was a bad paste, and I wanted to avoid including the invitee's public key because then we'd have to get into talk about when a public key is generated vs when memproofs are requested. I decided to go with your suggestion but added a note about the race problem in https://github.com/holochain/docs-pages/pull/520/commits/5ca238acce48a98419adaa205d4f930a86123c60 .

src/pages/build/validation.md
pdaoust: yes -- unintentional chain forks -- but of course this is not something that app validation logic can catch. Is this too pedantic here?

src/pages/build/validate-callback.md
pdaoust: I wanted to avoid including the invitee's public key because then we'd have to get into talk about when a public key is generated vs when memproofs are requested. I decided to go with your suggestion but added a note about the race problem in https://github.com/holochain/docs-pages/pull/520/commits/5ca238acce48a98419adaa205d4f930a86123c60 .

src/pages/build/validate-callback.md
pdaoust:
```suggestion
This example checks that a movie's release date is within [sensible bounds](https://en.wikipedia.org/wiki/Roundhay_Garden_Scene).
```

Can we meet halfway?

src/pages/build/validation.md
pdaoust: It didn't read quite clearly enough to me; I edited your suggestion -- does this still seem clear to you?

src/pages/build/validation.md
pdaoust: It feels like important information. Maybe it's in the wrong spot?

pdaoust: Moved it out of here but added it elsewhere: https://github.com/holochain/docs-pages/pull/520/commits/34da873e10bb198abaababea0cf6274cc6d0e63f

src/pages/build/validation.md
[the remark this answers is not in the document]
pdaoust: > There's nothing stopping you mixing things up in app validation but it will slow down validation

Okay, I'm afraid to ask, in case you cook up something like [this](https://github.com/holochain/docs-pages/pull/520#issuecomment-2675264660) again... how can an earlier action know/include the hash of a later action in its data?

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
pdaoust: I think I'll stick with this stuff on a separate page with warnings about advanced-ness, based on others' feedback about ops themselves being advanced. Re: the stuff that's not clearly/correctly implemented, will this be okay? https://github.com/holochain/docs-pages/pull/520/commits/e20390ee2c52fac2dae2bb6d2dc82a99ae98daf3

src/pages/build/dht-operations.md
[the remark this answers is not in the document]
pdaoust: @ddd-mtl ah, that may be the problem. How about https://github.com/holochain/docs-pages/pull/520/commits/961ef7689c4a0fcd6c660b3356aa851613f614df

mattyg (dismissed):

pdaoust:
@mattyg I notice you approved this PR -- is that with or without resolution of the following?

* https://github.com/holochain/docs-pages/pull/520/files#r1968629985
* https://github.com/holochain/docs-pages/pull/520/files#r1968625910
* https://github.com/holochain/docs-pages/pull/520/files#r1968582357

src/pages/build/validate-callback.md
mattyg: I feel like linking to a wikipedia page about the first movie ever made just adds noise.

src/pages/build/validate-callback.md
mattyg: > a note about the race problem in https://github.com/holochain/docs-pages/commit/5ca238acce48a98419adaa205d4f930a86123c60 .

This is a lot of additional info. I would rather leave that out here and talk about it when we get to app patterns.

src/pages/build/validation.md
mattyg: The only time chain forks can occur unintentionally is during countersigning, right? So yes I would not mention it as countersigning is not stable.

mattyg:
> @mattyg I notice you approved this PR -- is that with or without resolution of the following?

Ah my bad I didn't click through all the "hidden" comments, I assumed it would show my own comments.

src/pages/build/validation.md
pdaoust: Oh, I guess countersigning too -- the classic case Art talks about, though, is losing your source chain and restoring from a backup that was missing the most recent records.

src/pages/build/validate-callback.md
pdaoust: yes, this pattern does introduce a deep rabbit hole that I'd rather not go down either. Maybe I'll revert this like you suggest.

src/pages/build/validation.md
mattyg: Hmm, well since validation treats it as malicious and we don't have a stable feature for reverting unintentional chain forks right now, I'd say leave it out.

Maybe it could be rephased as:

> Because every peer has the DNA's validation logic on their own machine and is expected to check the data they author before they publish it, **invalid data is treated as** an intentionally malicious act

src/pages/build/validate-callback.md
pdaoust: added a note about inductive validation, and resolved todos that reminded me to link to it: https://github.com/holochain/docs-pages/pull/520/commits/58e879857a5df243bf7e18d233c9ac944120d321

src/pages/build/validate-callback.md
pdaoust: Update: decided to keep the example -- it seems more readable somehow -- but removed the warning about deferred memproofs. https://github.com/holochain/docs-pages/pull/520/commits/138e890b7645bec86958c789d22fd131338aca60

src/pages/build/genesis-self-check-callback.md
pdaoust: I think the commits mentioned in https://github.com/holochain/docs-pages/pull/520/files/50efef1de4bf432a088105fe3760dd886b61efdf#r1968246956 also resolve this.

src/pages/build/validation.md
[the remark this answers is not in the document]
ThetaSinner: Oh nothing quite so nasty in this case. Within a zome call, you might create multiple things. You could compute an entry hash and create a link to it, then do the create that brings the entry into existence. That is atomic from the point of view of the caller and the actions still appear in the right order to an agent activity authority but somebody validating that this particular link type points to entries is likely going to have do multiple passes to resolve this.

ThetaSinner (approved):
Changes looking good! Left one thread open because it was a question but please feel free to resolve it

src/pages/build/validation.md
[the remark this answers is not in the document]
pdaoust: oh yeah, guess you could link to an uncreated entry -- I was thinking linking to an uncreated action, which seemed impossible :)

ThetaSinner (approved):

PR #526, opened 2025-02-07
Build guide: Connecting a front end

### LINKED ISSUE #523: Build guide: Connect from a front end
Show how to establish and reuse an app websocket connection from a JavaScript front end. Assume that:

1. The front end will be served up by a runtime that uses the `APP_INTERFACE_PORT` connection string injection technique and handles cap grant creation for you (i.e., the default situation when you use the scaffolding tool.
2. We don't know what front end framework the dev will be using (i.e., use vanilla JS)

pdaoust:
Depends on #527. Closes #523.

The example is meant to be as vanilla JS and as useful as possible -- a reusable function that creates an active connection or returns an existing one. LMK if it could be even more minimal.

### DIFF
--- a/.cspell/words-that-should-exist.txt
+++ b/.cspell/words-that-should-exist.txt
@@ -25,6 +25,7 @@ runtimes
 sandboxed
 sandboxing
 scaffolder
+snapshotted
 spacebar
 todo
 todos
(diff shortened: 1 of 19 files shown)


pdaoust:
Closing because somehow this work sneaked into #527

PR #527, opened 2025-02-07
Build Guide: Connecting Everything Together

pdaoust:
This is a container/overview page for dumping the following concepts:

* Connecting a front end #523 / #526 
* Calling a zome function #524 
* Capabilities #524 
* Signals #471

PR #528, opened 2025-02-07
Build guide: Calling zome functions

### LINKED ISSUE #492: Zome calls 2: Calling from other stuff
* Remote calls
* Cross-zome/cell calls
* front end
* Capabilities
* writes can fail
    * validation failure or source chain head moved
    * use `post_commit` for follow-up

Broken out from #469.

pdaoust:
Depends on #527 . Closes #492 . ~~Note that code hasn't been tested yet.~~ EDIT: all TS and Rust code compiles now.


src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+
+::: intro
+You can call a zome function from [various local and remote processes](/build/connecting-the-parts/#what-processes-can-connect-to-a-happ) and secure them with capability tokens, letting you implement secure interop easily.
+:::
+
+## Call a zome function from a front end
+
+To call a zome function, use [`AppWebsocket.prototype.callZome`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.callzome.md). It takes a [`CallZomeRequest`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.callzomerequest.md) and an optional timeout.
+
+This example gets all the movies by a given director, then logs their titles to the console. You'll need the [`getHolochainClient` example](/build/connecting-a-front-end/#connect-to-a-happ-with-the-javascript-client) from the Connecting to a Front End page.
```
jost-s, line 13: Which "Connecting to a Front End page"? Can you link to it?

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+
+```typescript
+import { decodeHashFromBase64, EntryHash } from '@holochain/client';
+
+class Movie {
```
jost-s, line 18: I'd go with an `interface` instead of a `class`. I think that's more idiomatic, but I don't have evidence for it =)

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+        console.log(movie.title);
+    }
+}
+```
+
+The client handles errors (both `ExternResult::Err(_)` errors from the zome function and other errors from Holochain itself) and deserializes the MessagePack return value for you. It also takes care of capability security<!--TODO: link when written --> for you.
+
+## Call a zome function from another zome in the same hApp
+
+Coordinator zomes can't call each other's functions directly; instead they do it via the [`hdk::prelude::call`](https://docs.rs/hdk/latest/hdk/p2p/fn.call.html) host function.
```
jost-s, line 52: That looks direct to me.
```suggestion
Coordinator zomes can call each other's functions directly via the [`hdk::prelude::call`](https://docs.rs/hdk/latest/hdk/p2p/fn.call.html) host function.
```

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+    let response = call(
+        // This indicates that `bar` is in the same cell.
+        CallTargetCell::Local,
+        "movies",
+        "get_movies_for_director".into(),
+        // This is a capability secret -- we'll explain later why it's not
+        // needed.
+        None,
+        EntryHash::from_raw_36(vec![/* bytes of Sergio Leone director entry's hash */]),
+    )?;
```
jost-s, line 67:
```suggestion
fn get_movies_for_director(director: EntryHash) -> ExternResult<Vec<Movie>> {
    let response = call(
        // This indicates that `bar` is in the same cell.
        CallTargetCell::Local,
        "movies",
        "get_movies_for_director".into(),
        // This is a capability secret -- we'll explain later why it's not
        // needed.
        None,
        director,
    )?;
```

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+fn get_movies_by_sergio_leone_from_movies_cell() -> ExternResult<Vec<Movie>> {
+    let response = call(
+        // Address the cell by its role name.
+        // You can also address it by its cell ID.
+        CallTargetCell::OtherRole("movies".into()),
+        "movies",
+        "get_movies_for_director".into(),
+        None,
+        EntryHash::from_raw_36(vec![/* bytes of Sergio Leone director entry's hash */]),
+    )?;
```
jost-s, line 92:
```suggestion
fn get_movies_for_directory_from_movies_cell(director: EntryHash) -> ExternResult<Vec<Movie>> {
    let response = call(
        // Address the cell by its role name.
        // You can also address it by its cell ID.
        CallTargetCell::OtherRole("movies".into()),
        "movies",
        "get_movies_for_director".into(),
        None,
        director,
    )?;
```

jost-s (approved):
Looking great! I made suggestions along with the approval. When it comes to the zome function examples, taking the director as parameter seems more likely a use case and also removes the distracting hash encoding comment from the function body.

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+        payload: decodeHashFromBase64(" -- base64-encoded Sergio Leone hash -- ")
+    });
+    console.log("Got movies");
+    for (let movie in movies) {
+        console.log(movie.title);
+    }
+}
+```
+
+The client handles errors (both `ExternResult::Err(_)` errors from the zome function and other errors from Holochain itself) and deserializes the MessagePack return value for you. It also takes care of capability security<!--TODO: link when written --> for you.
```
matthme, line 48: > It also takes care of capability security

Is this referring to the client? What is meant exactly by "taking care of capability security" here, that it has the zome call signed by launcher/kangaroo/... with the author key under the hood?

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+    match response {
+        ZomeCallResponse::Ok(output) => output
+            .decode()
+            .map_err(|e| wasm_error!(e.to_string()))?,
+        _ => Err(wasm_error!("Something bad happened")),
+    }
+}
+```
+
+These cases don't need to worry about capability security either, because they're covered by a special grant called the [**author grant**](/concepts/8_calls_capabilities/#author-grant). It permits calls made by any caller with the same public key as the callee cell's owner.
```
matthme, line 102: I'm not sure this part is fully clear to me the way it's formulated. Is this saying that the call from zome A to zome B inside zome function X is allowed naturally as long the two zomes belong to the same cell (and thus same agent) and the call to zome function X is being made with a valid capability?

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+Holochain allows agents to add and remove coordinator zomes from their cells. This permits upgrading and customization. But it also means that the zomes and functions that you _think_ are on the other end might not actually be there.
+!!!
+
+This example calls a function _in the same coordinator zome_ (or at least one with the same name) in a remote agent's cell. It assumes that the remote agent has granted access to their `get_movies_by_director` function with an [**unrestricted grant**](/concepts/8_calls_capabilities/#unrestricted), which doesn't require a capability secret.
+
+```rust
+use hdk::prelude::*;
+use movies_types::*;
+
+fn get_movies_by_sergio_leone_remote() -> ExternResult<Vec<Movie>> {
```
matthme, line 118: potentially apply same changes here with entry hash as input as Jost suggested above for the other zome calls.

src/pages/build/calling-zome-functions.md
pdaoust: oh, the link is there in the paragraph -- I'll make it wrap all the pertinent text.

src/pages/build/calling-zome-functions.md
pdaoust: In many of the other code examples we've used literals and specific cases like "get everything by Sergio Leone". However, I think that's because it was bare code not wrapped in a function. I guess this *is* a function though -- happy to start wrapping more of this specific code in a function and making it generic if that seems better.

src/pages/build/calling-zome-functions.md
pdaoust: Right. I'll try to make that clearer without going into too much detail.

src/pages/build/calling-zome-functions.md
pdaoust: Right again, and that applies to within the same cell _and_ to other cells within the same hApp. I'll try to explain the author grant a little more.

src/pages/build/calling-zome-functions.md
matthme: great, looking clearer now :+1:

jost-s (approved):
Approving again as a way of concluding my re-review. I find the function like this more realistic.

src/pages/build/calling-zome-functions.md
```diff
@@ -0,0 +1,149 @@
...
+    let bob_public_key = AgentPubKey::from_raw_36(vec![/* bytes of remote agent's key */]);
```
matthme: To be picky, I don't think that a remote agent's pubkey could be hard-coded that way here in practice because it would only be known at runtime so would need to be passed as a function argument. I'm not sure that that's relevant for the purpose of the docs though...

matthme (dismissed):
Found another nit-pick but other than that looking good and may also be ignored (feel free to resolve the conversation on my behalf if the change is not deemed necessary) :+1:

src/pages/build/calling-zome-functions.md
matthme: There's also the additional complexity in practice about actually getting that pubkey from somewhere. A typical pattern that works good for small DHTs at least is that every agent links their pubkey to an `ALL_AGENTS` anchor as part of `init()`.

matthme: But that's probably beyond the scope of this guide.

pdaoust: :) I accidentally resolved this as part of resolving the suggestion to make the functions generic :)

matthme (approved):

PR #532, opened 2025-02-13
Build guide: Capabilities

### LINKED ISSUE #524: Build guide: call a zome function / capabilities
Show:

* how to call a zome function from JS, using the code and assumptions from #523 
* how to call from another zome/cell in the same hApp (can move code from the Zome Functions page)
* how to set up capabilities

pdaoust:
Closes #524 . I'm interested to hear what other people think about this -- there are so many moving parts involved in making a minimal functional capability setup in a hApp that I found the code samples are...... large. I tried to pare them down to the essentials (but still, there's a lot of LOC just to retrieve and exercise a claim), and added a section at the end that integrates it all together into a working system.

It was even worse than this before; this _is_ a rewrite. Any suggestions for improvement/simplification of the code samples are welcome.


src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
+---
+title: "Capabilities"
+---
+
+::: intro
+Access to zome functions is secured by a variant of **capability-based security** that adds the ability to restrict access to a given set of callers, identified and authenticated by their public/private key pair.
```
matthme, line 6: Actually, there's also the "Transferable" option of having it only secret based, i.e. not tied to a public key. But maybe that's too much info for an intro?

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
...
+An agent generates a capability by storing a [`CapGrantEntry`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/struct.CapGrantEntry.html) system entry on their source chain using the [`create_cap_grant`](https://docs.rs/hdk/latest/hdk/capability/fn.create_cap_grant.html) host function.
+
+!!! Capabilities have to be created in every cell
+
+A cell's zome functions aren't accessible to anyone except the author until the agent creates capability grants for them. Capabilities only cover one cell in a hApp.
+!!!
+
+### Unrestricted
+
+In a hApp where agents can call each other's zome functions, <!--TODO: expand this to other cells on agent's device when `UseExisting` provisioning strategy is implemented-->it's usually necessary to create an unrestricted grant for at least one zome function that allows an agent to ask another agent for more capabilities.
```
matthme, line 34: > it's usually necessary to create an unrestricted grant for at least one zome function that allows an agent to ask another agent for more capabilities

I think I disagree with this statement. This assumes that the granting of additional cap grants is at least mediated by remote calls but I don't think that's at all necessary and maybe not even common. Intuitively I'd rather assume that there are coordinator functions to grant access to certain agents for certain functions for example, and a user can pick in the happ UI an agent key from a profile dropdown for example and grant that agent access to something. For that to work there are no prior remote calls involved at all...

matthme: I would probably rather frame it here that there are certain types of remote calls that, depending on the happ, may want to be accessible by default because they provide an important functionality of the app.

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
...
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn init() -> ExternResult<InitCallbackResult> {
+    let mut functions = BTreeSet::new();
+    functions.insert((zome_info()?.name, "recv_remote_signal".into()));
+    create_cap_grant(CapGrantEntry {
+        tag: "".into(),
```
matthme, line 46:
```suggestion
        tag: "remote signals".into(),
```

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
...
+        access: CapAccess::Unrestricted,
+        functions: GrantedFunctions::Listed(functions),
+    })?;
+    Ok(InitCallbackResult::Pass)
+}
+```
+
+### Transferrable
+
+Sometimes it doesn't matter who's calling a zome function, as long as they can supply the right secret. This is useful when there's an open number of bots, system services, or other agents that should be authorized to call the function.
```
matthme, line 56: > an open number of bots, system services, or other agents

Interesting, do you see use cases for this primarily for non-human agents? It sounds like it from the formulation...but I haven't actually ever thought too much about use cases in which transferable secrets are useful.

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
...
+            assignees,
+        },
+        functions,
+    };
+    create_cap_grant(cap_grant)?;
+
+    // Rather than returning the secret to Bob's UI so he can send it to her
+    // out-of-band, let's send it to her directly. In a more robust app, we'd
+    // want to check for failed delivery and set up a retry handler.
+    send_remote_signal(
```
matthme, line 336: This is missing a vector of agent keys I think which should be constructed from the requestor field of the `DelegateAuthorRequest` input.

src/pages/build/capabilities.md
matthme: So potentially a way to simplify/shorten the code examples would be to remove the remote call part of requesting a cap grant and assume that this needs to happen out of band which in the case of ghost writing also doesn't necessarily seem unplausible.

matthme: But I'm not implying that I think the code example is not good the way it is right now...I'm just mentioning that thought since you're asking about simplification in the PR description.

src/pages/build/capabilities.md
pdaoust: yes, what I'm saying here is that traditional cap security doesn't have that concept so ours adds it. Doesn't seem clear to me either, on a re-read.

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,448 @@
...
+title: "Capabilities"
+---
+
+::: intro
+Access to zome functions is secured by a variant of **capability-based security** that adds the ability to restrict access to a given set of callers, identified and authenticated by their public/private key pair.
+:::
+
+## Capability-based security, updated for agent-centric applications
+
+Traditional [capability-based security](https://en.wikipedia.org/wiki/Capability-based_security) works on a simple concept: the owner of a resource grants access to other processes by giving out a handle to the resource rather than direct access to it. (Usually in a client/server system, the handle is an authorization secret such as an [OAuth2 token](https://auth0.com/intro-to-iam/what-is-oauth-2).) Thus they can control the way the resource is used without needing to deal with access control lists or other access control methods. When the owner no longer wants the process to access the resource, they destroy the handle.
```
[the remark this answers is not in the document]
pdaoust: I like nit-picks like this. Thanks!

src/pages/build/capabilities.md
pdaoust: Useful question -- I guess I can see it'd be useful for human agents with multiple devices. Any other use cases?

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
+---
+title: "Capabilities"
+---
+
+::: intro
+Access to zome functions is secured by **capability-based security**. Holochain extends this concept by adding the ability to restrict access to a given set of callers.
```
ThetaSinner, line 6: I think the way this is implemented in Holochain is thoroughly confused (i.e. the implementers made things unclear). But I think the way this is being introduced here makes things more confusing rather than less.

What's going on here is that generally a capability is hard to revoke. In the example of OAuth2 the revocation is generally done based on time. The holder can destroy their own token but the issuer can't do anything. If you want to be able to revoke on demand rather than after a time interval, then you need to record what access tokens you have issued somehow.

The thing we call a `CapSecret` is just not a secret. It's an identifier for the capability. In the case of an assigned capability, it's completely safe to share. You could call it a secret in the transferable case but really it's used to identify the capability that the caller is wanting to use and access is granted to anyone. It's pretty hard to call a shared password a secret.

So the "unforgeable token" part comes from combining the capability identifier (`CapSecret`) with a signature from the agent's private key. Although it's actually the zome call that you sign, the message you send is unforgeable.

So in reality, this is really very close to capability-based access control. You can think of the `CapSecret` as the capability being granted. The capability has to be dynamic because you could permit access to any combination of your zome functions. Trying to name these things uniquely could get very confusing for the user. What we're adding on top of the simplest version of this is revocation. We can delete a capability whenever we want and no use of the "secret" or a valid agent signature will let you make calls.

Every time I look at this, I think that it looks more like an ACL with a pointless secret. If we renamed `CapSecret` to `CapabilityId` I think that would be a good start. Then we should document this as capability-based access control with revocation.

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+        grantor: input.grantor,
+        secret: input.secret,
+    };
+    create_cap_claim(cap_claim)
+}
+```
+
+## Use a capability secret
+
+To exercise a capability they've been granted, the agent needs to retrieve the claim from their source chain using the [`query`](https://docs.rs/hdk/latest/hdk/chain/fn.query.html) host function and supply the secret along with the zome call.
```
ThetaSinner, line 181: The only tests for this in Holochain don't do that. They store the secret in memory during the test. That is very poor testing! Not even one example of how this would actually be used.

I'm not unhappy with this being documented this way. It's valid, it works. It's not great though. There's no reason you should have to load all your cap claims to find one that could easily be identified by a SQL query that returns a single row.

This should be a feature request against the HDK please.

pdaoust: agreed, way ahead of you :) https://github.com/holochain/holochain/issues/4708

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+                ZomeCallResponse::Unauthorized(_, _, _, _, _) => Err(wasm_error!("Agent {} won't let us create a movie through them", claim.grantor)),
+                _ => Err(wasm_error!("Something unexpected happened")),
+            }
+        }
+        _ => Err(wasm_error!("Couldn't find delegate author capability for {}", author_to_delegate_to)),
+    }
+}
+```
+
+## Putting it all together
```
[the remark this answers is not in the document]
pdaoust: speaks to a growing concern I had last Friday; guess I hung onto it because something something 'sunk cost fallacy'. Removed in https://github.com/holochain/docs-pages/pull/532/commits/7e5a96ea704387aa8af29a755724057f7b200a91

src/pages/build/capabilities.md
pdaoust: fixed in https://github.com/holochain/docs-pages/pull/532/commits/d1c7caebee9d85d7d53a747a690b5b47f38bbeff#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR365-R366

src/pages/build/capabilities.md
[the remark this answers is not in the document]
pdaoust: https://github.com/holochain/docs-pages/pull/532/commits/d1c7caebee9d85d7d53a747a690b5b47f38bbeff#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR11

src/pages/build/capabilities.md
pdaoust: After this feedback I had a nagging feeling the comprehensive example really wasn't necessary -- not if we're not writing about patterns in this first pass of the Build Guide -- and ThetaSinner confirmed. After moving it around I just ended up deleting it: https://github.com/holochain/docs-pages/pull/532/commits/7e5a96ea704387aa8af29a755724057f7b200a91

src/pages/build/capabilities.md
pdaoust: https://github.com/holochain/docs-pages/pull/532/commits/d1c7caebee9d85d7d53a747a690b5b47f38bbeff#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR46

src/pages/build/capabilities.md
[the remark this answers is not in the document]
pdaoust: https://github.com/holochain/docs-pages/pull/532/commits/d1c7caebee9d85d7d53a747a690b5b47f38bbeff#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR11

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+
+* `CapAccess::Unrestricted`: any signing key can access the function(s) covered by the capability.
+* `CapAccess::Transferable`: any caller who possesses the secret can access the function(s). This is identical to traditional capability-based security.
+* `CapAccess::Assigned`: a caller must possess the secret _and_ sign the call with a known key.
+
+There's a fourth kind of capability, called the **author grant**, which covers any call made by a caller with the same key as the cell's agent ID --- that is, _the agent who owns the cell_. It's essentially a combination of unrestricted (in terms of what functions can be called) plus assigned (in terms of who may call those functions).
+
+All zome-to-zome calls within a cell and cell-to-cell calls within an agent's hApp instance are covered by the author grant. UIs are also covered by the author grant if they're hosted by one of the [well-known Holochain runtimes](/build/happs/#package-a-happ-for-distribution).
```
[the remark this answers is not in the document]
pdaoust: What about treating it as a given? As in, we assume that you're scaffolding a hApp and hence will be using one of our runtimes, so you don't need to worry about provisioning grants for the UI? In that case we don't nee to say anything about the runtime here -- just say "this also applies to a UI that's bundled with a hApp".

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Create a capability grant
+
+An agent generates a capability by storing a [`CapGrantEntry`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/struct.CapGrantEntry.html) system entry on their source chain using the [`create_cap_grant`](https://docs.rs/hdk/latest/hdk/capability/fn.create_cap_grant.html) host function.
+
+!!! Capabilities have to be created in every cell
```
mattyg, line 27:
```suggestion
!!! note Capabilities have to be created in every cell
```

src/pages/build/dnas.md
```diff
@@ -138,7 +138,7 @@ So Holochain manages the dependency mapping for you, allowing you to write code
 
 When do you decide whether a hApp should have more than one DNA? Whenever it makes sense to have multiple separate networks or databases within the hApp. These are the most common use cases:
 
-* **Dividing responsibilities.** For instance, a video sharing hApp may have one group of peers who are willing to index video metadata and offer search services and another group of peers who are willing to host and serve videos, along with people who just want to watch them. This DNA could have `search` and `storage` DNAs, along with a main DNA that allows video watchers to look up peers that are offering services and query them.
+* **Dividing responsibilities.** For instance, a video sharing hApp may have one group of peers who are willing to index video metadata and offer search services and another group of peers who are willing to host and serve videos, along with people who just want to watch them. This DNA could have `search` and `storage` DNAs, along with a main DNA that allows video watchers to look up peers that are offering services and query them. {#dividing-responsibilities}
```
mattyg, line 141: Doesn't seem like this is linked to anywhere?

src/pages/build/capabilities.md
pdaoust: Trying to wrap my head around this in light of what I know about ocap systems. Are you saying that `Transferable` lacks the property of confinement, whereas 'Assigned' gets you what [Demolishing Capability Myths](https://srl.cs.jhu.edu/pubs/SRL2003-02.pdf) calls 'Property F: Access-controlled delegation channels' (as I understand it, in Holochain terms that means that in order for Alice to give to Bob a capability she received from Carol, Alice has to give Bob a capability to one of her own zome functions that then exercises the Carol capability)?

If so, then I get what you're talking about -- for Transferable, the secret looks more like an API token, security through obfuscation, and for Assigned, the secret is unnecessary but useful as an identifier for later exercise.

My question is, what's the next steps? Do we talk about this in the guide (or alternatively update the core concepts)? If the problem is in the design, I feel like the most we can do in the guide and the CC is to put up warnings that X isn't secure, Y is, to help knowledgeable people map their understanding of ocaps to what Holochain does and doesn't do.

src/pages/build/capabilities.md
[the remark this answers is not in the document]
ThetaSinner: Good with me yes, just don't say anything and people shouldn't worry about why it works until they try to build their own runtime or use the client libraries outside of our tooling.

src/pages/build/capabilities.md
ThetaSinner: Nice :) I remember seeing that issue now you say!

src/pages/build/capabilities.md
pdaoust: > I don't agree, an API token [...] is supposed to uniquely identify its user.

Ahh, is that so. Shows how much I integrated external APIs in my last job 😅

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+    };
+    create_cap_grant(cap_grant)?;
+
+    Ok(secret)
+}
+```
+
+## Store a capability claim
+
+Once an has gotten a capability secret from a grantor, they need to store it as a [`CapClaim`](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/struct.CapClaim.html) entry with the [`create_cap_claim`](https://docs.rs/hdk/latest/hdk/capability/fn.create_cap_claim.html) host function so they can use it later when they want to call the functions they've been granted access to.
```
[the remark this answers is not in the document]
pdaoust: Pains me to say this, but I'm standardising on US English in the documentation.

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+
+Holochain extends this concept for zome calls, first by requiring that the payload of every call be signed by a private key. Let's take a look at the [`CapAccess` enum](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/enum.CapAccess.html) which defines the kinds of **capability grant** you can use in your hApp:
+
+* `CapAccess::Unrestricted`: any signing key can access the function(s) covered by the capability.
+* `CapAccess::Transferable`: any caller who possesses the secret can access the function(s). This is identical to traditional capability-based security.
```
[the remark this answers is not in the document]
pdaoust: rewrote this whole section: https://github.com/holochain/docs-pages/pull/532/commits/875766ae9b9dc8a2f6da135893581e5e1519ca30#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR6-R15

src/pages/build/capabilities.md
pdaoust: https://github.com/holochain/docs-pages/pull/532/commits/1d3b69c7d42a086c3997a8383b9031cac48b4785

Admonitions as implemented in this site are limited -- you only get one type, and any attempt to use any other results in the same styling. So I just use `info` for everything.

src/pages/build/capabilities.md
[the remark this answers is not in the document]
pdaoust: simplified in https://github.com/holochain/docs-pages/pull/532/commits/26a6b7d94b57f9ba90d262fd1df1bb9935ec7161#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR19

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+* `CapAccess::Assigned`: a caller must possess the secret _and_ sign the call with a known key.
+
+There's a fourth kind of capability, called the **author grant**, which covers any call made by a caller with the same key as the cell's agent ID --- that is, _the agent who owns the cell_. It's essentially a combination of unrestricted (in terms of what functions can be called) plus assigned (in terms of who may call those functions).
```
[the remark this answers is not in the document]
pdaoust: simplified in https://github.com/holochain/docs-pages/commit/26a6b7d94b57f9ba90d262fd1df1bb9935ec7161#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR19

src/pages/build/capabilities.md
```diff
@@ -0,0 +1,479 @@
...
+    create_cap_grant(cap_grant)?;
+
+    // Return the secret so the grantor can send it to the requestor.
+    Ok(secret)
+}
+```
+
+### Assigned
+
+If you're concerned about capability secrets being leaked, you can bind a secret to one or more public keys. The zome call's provenance must match one of these public keys, and the payload signature must be valid for the provenance. This function rewrites `approve_delegate_author_request` to create an assigned grant.
```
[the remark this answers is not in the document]
pdaoust: does this help? https://github.com/holochain/docs-pages/pull/532/commits/875766ae9b9dc8a2f6da135893581e5e1519ca30#diff-7b1eadffe9e8704c4c3727facf77b0cfc8a1271d48504716c6e31f637086247dR87

src/pages/build/dnas.md
pdaoust: casualty of an edit. fixed in https://github.com/holochain/docs-pages/pull/532/commits/5fc01ab77b45214b40ed0f99f30e45722f8bdc94

ThetaSinner (dismissed):
Nice, I think this reads well! I get what I need to know and I'm not left with questions I couldn't answer by following links. Ideal :)

pdaoust:
@ThetaSinner can I mark your conversations as resolved then? 🤞🏼 🤞🏼 🤞🏼

src/pages/build/capabilities.md
ThetaSinner: Yes, that looks good!

src/pages/build/capabilities.md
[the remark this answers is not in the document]
ThetaSinner: Yes, clear

ThetaSinner (approved):
Looking good!

PR #534, opened 2025-02-18
Build guide: signals

### LINKED ISSUE #471: Practical docs for using Signals
Must include:
- code examples
- gotchas

* Signals
    * Local signals to UI
    * Remote signals to peers on same DHT
        * needs cap grant on `recv_remote_signal`
        * (future) pub/sub to replace most use cases for remote signals
    * not supported between zomes or between cells :(

pdaoust:
Closes #471 . Covers both local and remote signals.

### DIFF
--- a/src/pages/build/signals.md
+++ b/src/pages/build/signals.md
@@ -152,7 +152,7 @@ fn send_heartbeat_via_remote_call(agent: AgentPubKey) -> ExternResult<()> {
 }
 ```
 
-This means an agent needs to set up an [`Unrestricted` capability grant](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/enum.CapAccess.html#variant.Unrestricted)<!--TODO: link to capabilities page --> for it, so other agents can call it. Take care that this function does as little as possible, to avoid people abusing it. Permissions and privileges are another topic which we'll talk about soon.<!-- TODO: delete this sentence -->
+Take care that `recv_remote_signal` does as little as possible, to avoid people abusing it. Permissions and privileges are another topic which we'll talk about soon.<!-- TODO: delete this sentence and link to capabilities page -->
 
 It also means that `send_remote_signal` always routes the call to a coordinator zome of the same name as the caller. Because [the remote agent might map that name to a different coordinator zome, or no zome at all](/build/calling-zome-functions/#remote-call-unknown-routing), this function might be handled in unexpected ways on the receiver's end.
 
--- a/src/pages/build/signals.md
+++ b/src/pages/build/signals.md
@@ -43,7 +43,7 @@ pub fn post_commit(committed_actions: Vec<SignedActionHashed>) {
 
 ### Listen for a signal
 
-The UI subscribes to signals with the [`AppWebsocket.prototype.on`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.on.md) method. The signal handler should expect signals from any coordinator zome in any cell in the agent's hApp instance, and can discriminate between them by cell ID and zome name.
+Holochain emits local signals over active app WebSocket connections, and a client should provide a way to receive these signals. For instance, with the TypeScript client, you can subscribe to signals with the [`AppWebsocket.prototype.on`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.on.md) method. The signal handler should expect signals from _any coordinator zome in any cell_ in the agent's hApp instance, and should discriminate between them by cell ID and zome name.
 
 <!-- FIXME(0.5): does SignalType still exist? -->
 
--- a/src/pages/build/signals.md
+++ b/src/pages/build/signals.md
@@ -37,56 +37,6 @@ pub fn post_commit(committed_actions: Vec<SignedActionHashed>) {
 }
 ```
 
-This example shows a 'heartbeat' feature, where peers can periodically ping each other to let them know they're still online.
-
-```rust
-use hdk::prelude::*;
-
-#[derive(Serialize, Deserialize, Debug)]
-#[serde(tag = "type", content = "value", rename_all = "snake_case")]
-pub enum LocalSignal {
-    Heartbeat(AgentPubKey),
-}
-
-#[hdk_extern]
-pub fn init() -> ExternResult<InitCallbackResult> {
-    // Let all agents send heartbeat messages to each other.
-    let mut fns = BTreeSet::new();
-    fns.insert((zome_info()?.name, "receive_heartbeat".into()));
-    create_cap_grant(CapGrantEntry {
-        tag: "heartbeat".into(),
-        access: CapAccess::Unrestricted,
-        functions: GrantedFunctions::Listed(fns),
-    })?;
-    Ok(InitCallbackResult::Pass)
-}
-
-// An agent's UI calls this function periodically to let others know they're
-// online.
-#[hdk_extern]
-pub fn send_heartbeat(receivers: Vec<AgentPubKey>) -> ExternResult<()> {
-    for agent in receivers {
-        call_remote(
-            agent,
-            zome_info()?.name,
-            "receive_heartbeat".into(),
-            None,
-            (),
-        )?;
-    }
-    Ok(())
-}
-
-#[hdk_extern]
-pub fn receive_heartbeat() -> ExternResult<()> {
-    // Find out who called our heartbeat function.
-    let caller = call_info()?.provenance;
-    // Tell the UI that the caller is still online.
-    emit_signal(LocalSignal::Heartbeat(caller))?;
-    Ok(())
-}
-```
-
 ### Listen for a signal
 
 The UI subscribes to signals with the [`AppWebsocket.prototype.on`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.on.md) method. The signal handler should expect signals from any coordinator zome in any cell in the agent's hApp instance, and can discriminate between them by cell ID and zome name.
@@ -97,7 +47,6 @@ import { SignalType, encodeHashToBase64 } from "@holochain/client";
 
 // Duplicate your zome's signal types in the UI.
 type MyZomeSignal =
-    | { type: "heartbeat"; value: AgentPubKey }
     | { type: "action_written"; value: ActionHash };
 
 // Use the connection establishment function from
@@ -117,9 +66,6 @@ getHolochainClient().then(client => {
 
         const payload: MyZomeSignal = appSignal.payload;
         switch (appSignal.payload.type) {
-            case "heartbeat":
-                console.log(`agent ${encodeHashToBase64(payload.value)} is still online`);
-                break;
             case "action_written":
                 console.log(`action hash ${encodeHashToBase64(payload.value)} written`);
         }
@@ -131,7 +77,7 @@ getHolochainClient().then(client => {
 
 Agents can also send remote signals to each other using the [`send_remote_signal`](https://docs.rs/hdk/latest/hdk/p2p/fn.send_remote_signal.html) host function and a [`recv_remote_signal` callback](/build/callbacks-and-lifecycle-hooks/#define-a-recv-remote-signal-callback), which takes a single argument of any type and returns `ExternResult<()>`.
 
-This example rewrites and expands the `heartbeat` function above to use remote signals.
+This example implements a 'heartbeat' feature, where peers can periodically ping each other to let them know they're still online.
 
 ```rust
 use hdk::prelude::*;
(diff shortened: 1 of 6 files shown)


src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+    Heartbeat(AgentPubKey),
+}
+
+#[hdk_extern]
+pub fn init() -> ExternResult<InitCallbackResult> {
+    // Let all agents send heartbeat messages to each other.
+    let mut fns = BTreeSet::new();
+    fns.insert((zome_info()?.name, "receive_heartbeat".into()));
+    create_cap_grant(CapGrantEntry {
+        tag: "".into(),
```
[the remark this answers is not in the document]
matthme: I imagine that an empty tag is not good practice since cap grants may also be created outside init and an empty tag is probably prone to collisions.

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+pub fn post_commit(committed_actions: Vec<SignedActionHashed>) {
+    // Tell the UI about every action that any function in this zome has
+    // written.
+    for action in committed_actions {
+        let _ = emit_signal(LocalSignal::ActionWritten(action.action_address()));
+    }
+}
+```
+
+This example shows a 'heartbeat' feature, where peers can periodically ping each other to let them know they're still online.
```
matthme, line 40: I think typically I would implement such a feature using `send_remote_signal` rather than a remote call. I have never tried it with a remote call actually but I'm wondering whether you'd run into issues with blocking other calls because remote calls waits 60 seconds before they time out and I don't know what's the behavior if a peer is offline in that regard, which I imagine would be a common scenario.
So I'm wondering whether we should at least write something about that here to not imply the following code to be "best practice" for a heartbeat implementation.

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+    | { type: "action_written"; value: ActionHash };
+
+// Use the connection establishment function from
+// https://developer.holochain.org/build/connecting-a-front-end/#connect-to-a-happ-with-the-javascript-client
+getHolochainClient().then(client => {
+    // Subscribe to signals.
+    client.on("signal", (signal: Signal) => {
+        // There's currently only one useful signal type to listen for -- an
+        // app signal.
+        if (!(SignalType.App in signal)) return;
```
matthme, line 110: This is about to change in holochain 0.5 with the new enum serialization. Just as a heads up. It will be something like
```
if (signal.type !== "app") return;
```

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+                console.log(`action hash ${encodeHashToBase64(payload.value)} written`);
+        }
+    });
+});
+```
+
+## Remote signals
+
+Agents can also send remote signals to each other using the [`send_remote_signal`](https://docs.rs/hdk/latest/hdk/p2p/fn.send_remote_signal.html) host function and a [`recv_remote_signal` callback](/build/callbacks-and-lifecycle-hooks/#define-a-recv-remote-signal-callback), which takes a single argument of any type and returns `ExternResult<()>`.
+
```
matthme, line 133: Maybe it would be worth writing here about the distinction from a remote call, i.e. that these remote signals are not waiting for a response whatsoever and have no guarantee of delivery.

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn init(_: ()) -> ExternResult<InitCallbackResult> {
+    let mut fns = BTreeSet::new();
+    // Open up access for the remote signal handler callback to everyone on
+    // the network -- see the note after this example.
+    fns.insert((zome_info()?.name, "recv_remote_signal".into()));
+    create_cap_grant(CapGrantEntry {
+        tag: "".into(),
```
matthme, line 146:
```suggestion
        tag: "remote signals".into(),
```

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+    send_remote_signal(
+        RemoteSignal::Heartbeat,
+    )
+}
+
+#[hdk_extern]
+pub fn recv_remote_signal(payload: RemoteSignal) -> ExternResult<()> {
+    if let RemoteSignal::Heartbeat = payload {
+        let caller = call_info()?.provenance;
+        emit_signal(LocalSignal::Heartbeat(caller))?;
```
matthme, line 174:
```suggestion
        // On the receiving end we forward the remote signal to the front-end by emitting a local signal
        emit_signal(LocalSignal::Heartbeat(caller))?;
```

src/pages/build/signals.md
```diff
@@ -0,0 +1,216 @@
...
+    }
+    Ok(())
+}
+```
+
+!!! info Remote signal handlers are just zome functions
+`send_remote_signal` is sugar for a [remote call](/build/calling-zome-functions/#call-a-zome-function-from-another-agent-in-the-network) to a zome function named  `recv_remote_signal` with no capability secret<!-- TODO: link to capabilities page -->. It works differently from a usual remote call, though, in that it's 'send-and-forget' --- it won't return an error if anything fails. In practice, these two are equivalent:
```
matthme, line 181: > `send_remote_signal` is sugar for a remote call to a zome function named  `recv_remote_signal` with no capability secret

Is this really correct? Because as written above, there is no timeout involved so I'm wondering whether there is actually a difference of how it's implemented under the hood but I'm not familiar with that part of the holochain code.

matthme: So I'm actually sceptical that they are equivalent in practice. Have you tried that out, including situations where peers to which a heartbeat is sent are offline?

pdaoust: hmmmm, I forgot that there's a timeout that gets ignored at the core/ribosome level. Good point.

src/pages/build/signals.md
```diff
@@ -0,0 +1,166 @@
...
+**Signals** are messages emitted by coordinator zomes, either locally to a front end or remotely to another agent cell in a DNA's network. They help you automate processes in your application and make it dynamic and responsive.
+:::
+
+## Local signals
+
+**Local signals** are sent to [front ends](/build/connecting-a-front-end/) listening on the agent's local machine.
+
+### Emit a signal
+
+Your coordinator zome emits a signal with the [`emit_signal`](https://docs.rs/hdk/latest/hdk/p2p/fn.emit_signal.html) host function. You can call this function from a regular [zome function](/build/zome-functions/) or the [`init`](/build/callbacks-and-lifecycle-hooks/#define-an-init-callback), [`recv_remote_signal`](/build/callbacks-and-lifecycle-hooks/#define-a-recv-remote-signal-callback), or [`post_commit`](/build/callbacks-and-lifecycle-hooks/#define-a-post-commit-callback) callbacks.
```
ThetaSinner, line 15:
```suggestion
Your coordinator zome emits a signal with the [`emit_signal`](https://docs.rs/hdk/latest/hdk/p2p/fn.emit_signal.html) host function. It takes any serializable input and you can call this function from a regular [zome function](/build/zome-functions/) or the [`init`](/build/callbacks-and-lifecycle-hooks/#define-an-init-callback), [`recv_remote_signal`](/build/callbacks-and-lifecycle-hooks/#define-a-recv-remote-signal-callback), or [`post_commit`](/build/callbacks-and-lifecycle-hooks/#define-a-post-commit-callback) callbacks.
```

src/pages/build/signals.md
ThetaSinner: Yes agree that should be documented but the same is actually true of local signals. No guarantees in either case. I've commented similarly above.

src/pages/build/signals.md
```diff
@@ -0,0 +1,166 @@
...
+        zome_info()?.name,
+        "recv_remote_signal".into(),
+        None,
+        RemoteSignal::Heartbeat
+    )?;
+    Ok(())
+}
+```
+
+This means an agent needs to set up an [`Unrestricted` capability grant](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/capability/enum.CapAccess.html#variant.Unrestricted)<!--TODO: link to capabilities page --> for it, so other agents can call it. Take care that this function does as little as possible, to avoid people abusing it. Permissions and privileges are another topic which we'll talk about soon.<!-- TODO: delete this sentence -->
```
ThetaSinner, line 152: Oh okay, we're saying something about this in two places, maybe we can unify?

ThetaSinner (commented):
This is looking good. I have a few minor comments but mostly happy. I like how focused this one is on giving code examples and just enough description.

src/pages/build/signals.md
```diff
@@ -0,0 +1,166 @@
...
+// Again, it's good practice to define your remote signal type as an enum so
+// you can add more message types later.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "type", content = "value", rename_all = "snake_case")]
+enum RemoteSignal {
+    Heartbeat,
+}
+
+#[hdk_extern]
+pub fn send_heartbeat(receivers: Vec<AgentPubKey>) -> ExternResult<()> {
```
ThetaSinner, line 110: If we're documenting with code like this, is it explicitly worth calling out that this forces Holochain to open and maintain connections for the heartbeat? So you wouldn't want logic that connects to everyone who is using your app. Holochain isn't designed to operate that many connections. It's fine like we do for syn with 2-10 people updating stickies, but in general you shouldn't be deliberately opening connections to every agent on the network.

Maybe that's out of scope or maybe that's a motivation to keep code samples even more minimal? This could just be a message getting sent, rather than a pattern

pdaoust: https://github.com/holochain/docs-pages/pull/534/commits/fc2d085109c539fbd4492e653272d6abd4b701f3

src/pages/build/signals.md
pdaoust: https://github.com/holochain/docs-pages/pull/534/commits/b623e47ef3e559141e977fe8c4bbe8d7563f4baa

src/pages/build/signals.md
```diff
@@ -0,0 +1,166 @@
...
+type MyZomeSignal =
+    | { type: "action_written"; value: ActionHash };
+
+// Use the connection establishment function from
+// https://developer.holochain.org/build/connecting-a-front-end/#connect-to-a-happ-with-the-javascript-client
+getHolochainClient().then(client => {
+    // Subscribe to signals.
+    client.on("signal", (signal: Signal) => {
+        // There's currently only one useful signal type to listen for -- an
+        // app signal.
```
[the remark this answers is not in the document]
pdaoust: https://github.com/holochain/docs-pages/pull/534/commits/d2421fb100256998833555f9608b14f2022321e7

src/pages/build/signals.md
```diff
@@ -0,0 +1,166 @@
...
+    // written.
+    for action in committed_actions {
+        let _ = emit_signal(LocalSignal::ActionWritten(action.action_address()));
+    }
+}
+```
+
+### Listen for a signal
+
+The UI subscribes to signals with the [`AppWebsocket.prototype.on`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.on.md) method. The signal handler should expect signals from any coordinator zome in any cell in the agent's hApp instance, and can discriminate between them by cell ID and zome name.
```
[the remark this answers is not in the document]
pdaoust: https://github.com/holochain/docs-pages/pull/534/commits/fa1b8f7d0753d2583d592984e1d00a3339e8698b and https://github.com/holochain/docs-pages/pull/534/commits/5e746c562d693141d284ca33ee215f1be7f1fc0a

src/pages/build/signals.md
pdaoust: https://github.com/holochain/docs-pages/commit/fa1b8f7d0753d2583d592984e1d00a3339e8698b and https://github.com/holochain/docs-pages/commit/5e746c562d693141d284ca33ee215f1be7f1fc0a

src/pages/build/signals.md
matthme: The constant has been removed for now (arguably that was unnecessary), the values are lower case (snake case) instead of upper case and get moved into the `type` field instead of being at the top-level of the Signal object.

src/pages/build/signals.md
matthme: This one can be considered resolved on my end.

src/pages/build/signals.md
matthme: @pdaoust did you not see this one or think it's unnecessary?

matthme (approved):
One more open suggested change but I can live without it too.

ThetaSinner (dismissed):
Looking good!

src/pages/build/signals.md
pdaoust: neither -- forgot about it :)

matthme (approved):

PR #535, opened 2025-02-19
Build Guide: querying source chains

### LINKED ISSUE #474: Practical guide for using Source Chain Query
Must include:
- code examples
- gotchas

* Source chain
    * Querying
        * `query` for local state querying
    * anything else?

pdaoust:
Closes #474 . Talks about it from the perspective of oneself (`query`), another agent (`get_agent_activity`), and validation (`must_get_agent_activity`).

### DIFF
--- a/src/pages/build/querying-source-chains.md
+++ b/src/pages/build/querying-source-chains.md
@@ -1,0 +1,3 @@
+---
+title: "Querying Source Chains"
+---
--- a/src/pages/build/querying-source-chains.md
+++ b/src/pages/build/querying-source-chains.md
@@ -147,7 +147,7 @@ pub fn check_status_of_peer_source_chain(agent: AgentPubKey) -> ExternResult<Cha
 
 Validation imposes an extra constraint on source chain queries. A source chain can grow over time, including branching or forking. That means a source chain, when retrieved by agent public key alone, is a non-deterministic source of data, which you can't use in validation<!-- TODO: link to validation page -->. Instead, you can use [`must_get_agent_activity`](https://docs.rs/hdi/latest/hdi/chain/fn.must_get_agent_activity.html), whose [filter struct](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/chain/struct.ChainFilter.html) and return value remove non-determinism.
 
-`must_get_agent_activity` only allows you to select a contiguous, bounded slice of a source chain, and doesn't return any information about the validity of the actions in that slice or the chain as a whole.
+`must_get_agent_activity` only allows you to select a contiguous, bounded slice of a source chain, and doesn't return any information about the validity of the actions in that slice or the chain as a whole. It needs to get the entire slice from an authority, so it's best to use it only when validating a `RegisterAgentActivity` operation<!-- TODO: link to DHT ops page -->, where the validating authority already has that data.
 
 <!-- TODO: move this to validation page or must_get_* page -->
 
--- a/src/pages/build/querying-source-chains.md
+++ b/src/pages/build/querying-source-chains.md
@@ -19,7 +19,7 @@ Whether an agent is querying their own source chain or another agent's, you buil
 
 * <code>sequence_range: <a href="https://docs.rs/holochain_zome_types/latest/holochain_zome_types/query/enum.ChainQueryFilterRange.html">ChainQueryFilterRange</a></code>: A start and end point on the source chain, either:
     * `Unbounded`: the beginning and current tip of the chain
-    * `ActionSeqRange(u32, u32)`: start and end sequence indices, inclusive.
+    * `ActionSeqRange(u32, u32)`: start and end sequence indices, inclusive, zero-based.
     * `ActionHashRange(ActionHash, ActionHash)`: start and end action hashes, inclusive.
     * `ActionHashTerminated(ActionHash, u32)`: an action hash plus the _n_th actions preceding it.
 * <code>entry_type: Option&lt;Vec&lt;<a href="https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/action/enum.EntryType.html">EntryType</a>&gt;&gt;</code>: Only select the given entry types, which can include both system and app entry types.
--- a/src/pages/build/querying-source-chains.md
+++ b/src/pages/build/querying-source-chains.md
@@ -3,7 +3,7 @@ title: "Querying Source Chains"
 ---
 
 ::: intro
-An agent can **query their source chain** for a history of the [records](/build/working-with-data/#entries-actions-and-records-primary-data) they've written, including [**link**](/build/links/) and public and private [**entry**](/build/entries/) data, which includes **capability grants and claims**<!--TODO: link to capabilities -->. They can also query the public portion of another agent's source chain in a [**zome function**](/build/zome-functions/) or [**`validate` callback**](/build/callbacks-and-lifecycle-hooks/#define-a-validate-callback).
+An agent can **query their source chain** for a history of the [records](/build/working-with-data/#entries-actions-and-records-primary-data) they've written, including [**link**](/build/links-paths-and-anchors/) and public and private [**entry**](/build/entries/) data, which includes **capability grants and claims**<!--TODO: link to capabilities -->. They can also query the public portion of another agent's source chain in a [**zome function**](/build/zome-functions/) or [**`validate` callback**](/build/callbacks-and-lifecycle-hooks/#define-a-validate-callback).
 :::
 
 An agent's source chain is their record of local state changes. It's a multi-purpose data structure, and can be interpreted in different ways, including as:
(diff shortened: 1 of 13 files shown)


src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Query an agent's own source chain
+
+An agent can query their own source chain with the [`query`](https://docs.rs/hdk/latest/hdk/chain/fn.query.html) host function, which takes a `ChainQueryFilter` and returns a `Vec<Record>` wrapped in an `ExternResult`.
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn get_all_movies_i_authored() -> Vec<Record> {
```
ThetaSinner, line 75: I don't know how many people agree with me, we shouldn't return records from zome calls in general. I'm not super fussy about that appearing here because it's just demonstrating something simple and we don't need a bunch of type conversion, but I will call that out when I see it.

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+    query(ChainQueryFilter::new()
+        .entry_type(EntryType::App(UnitEntryTypes::Movie.into()))
+        .include_entries(true)
+    )
+}
+```
+
+## Query another agent's source chain
```
ThetaSinner, line 83: That's a really different topic. You don't get back their source chain in the same way you do locally and it's for quite different purposes. 

There's a huge downside, in that you have to fetch and cache their entire chain to do this. Plus it's really for examining an agent's state, not the content of their chain. You see what valid/invalid data they have, the top of their chain, any warrants etc. That's not the same as retrieving your own source chain. I think this would be better saved until we're dealing with warrants and blocking. 

This function was actually flagged to be renamed in the run-up to the Holo launch to make it clearer what it does.

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+            .include_entries(true),
+        ActivityRequest::Full
+    )?;
+
+    // Now try to retrieve the records for all of the action hashes, turning
+    // the whole result into an Err if any record retrieval returns an Err.
+    let chain_with_entries = activity.valid_activity
+        .iter()
+        .map(|a| {
+            let maybe_record = get(a.1, GetOptions::network())?;
```
ThetaSinner, line 107: This could issue thousands of `get`s and cause them all to be cached. This is so expensive, I'd call it an antipattern. Data that this agent intended to share should be discovered through the DHT with links and the like, ideally not via their chain unless you have a really good reason to audit their activity or something.

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+            Ok((a.1, maybe_record))
+        })
+        .collect();
+    Ok(chain_with_entries)
+}
+```
+
+### Get source chain status summary
+
+`get_agent_activity` is for more than just querying an agent's source chain. It also returns the _status of the agent's source chain_, which includes evidence of [source chain forks](/resources/glossary/#fork-source-chain) and invalid actions.
```
ThetaSinner, line 120: This is the primary intention of the function, and why I think it's probably not meant for this page.

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+* <code>Invalid(<a href="https://docs.rs/hdk/latest/hdk/prelude/struct.ChainHead.html">ChainHead</a>)</code>: An invalid record was found at the given sequence index and action hash.
+
+## Query another agent's source chain for validation
+
+Validation imposes an extra constraint on source chain queries. A source chain can grow over time, including branching or forking. That means a source chain, when retrieved by agent public key alone, is a non-deterministic source of data, which you can't use in validation<!-- TODO: link to validation page -->. Instead, you can use [`must_get_agent_activity`](https://docs.rs/hdi/latest/hdi/chain/fn.must_get_agent_activity.html), whose [filter struct](https://docs.rs/holochain_integrity_types/latest/holochain_integrity_types/chain/struct.ChainFilter.html) and return value remove non-determinism.
+
+`must_get_agent_activity` only allows you to select a contiguous, bounded slice of a source chain, and doesn't return any information about the validity of the actions in that slice or the chain as a whole.
```
ThetaSinner, line 150: Here would be a good place to suggest that this is only used within RegisterAgentActivity validation

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+    let movie_update_actions: Vec<Update> = filter_movie_update_actions
+        .filter_actions(result_as_actions)
+        .iter()
+        .map(|a| a.try_into().unwrap())
+        .collect();
+
+    // Now find out how many times we've tried to update the same `Movie` as
+    // we're trying to update now.
+    let times_I_updated_this_movie = movie_update_actions
+        .fold(1, |c, a| c + if a.original_action_address == action.original_action_address { 1 } else { 0 });
```
ThetaSinner, line 198: Why not `fold(0`?

ThetaSinner (commented):
The query and validation lookups look good. I'm suggesting keeping coordinator agent activity as a separate topic

src/pages/build/querying-source-chains.md
```diff
@@ -0,0 +1,228 @@
...
+* a store for private entry data, or
+* a store for capability grants and claims.
+
+## Filtering a query
+
+Whether an agent is querying their own source chain or another agent's, you build a query with the [`ChainQueryFilter`](https://docs.rs/holochain_zome_types/latest/holochain_zome_types/query/struct.ChainQueryFilter.html) struct, which has a few filter types:
+
+* <code>sequence_range: <a href="https://docs.rs/holochain_zome_types/latest/holochain_zome_types/query/enum.ChainQueryFilterRange.html">ChainQueryFilterRange</a></code>: A start and end point on the source chain, either:
+    * `Unbounded`: the beginning and current tip of the chain
+    * `ActionSeqRange(u32, u32)`: start and end sequence indices, inclusive.
```
mattyg, line 22:
```suggestion
    * `ActionSeqRange(u32, u32)`: start and end sequence indices, inclusive, zero-based.
```

src/pages/build/querying-source-chains.md
mattyg: Scaffolding currently returns Records from zome calls. An app's UI will generally want the author, action hash, and timestamp along with the entry data.

I see that Record is a cumbersome data structure, but unless we wanted to implement something like https://github.com/holochain-open-dev/infrastructure/blob/15f89a9a2026b4c80b092a55b2faed605f435b91/packages/utils/src/entry-record.ts#L10 in the hdk, and update scaffolding to use it, I'd suggest that returning Records is the best practice for now.

src/pages/build/querying-source-chains.md
pdaoust: cuz it's iterating over _previous_ actions not including this one, which is _also_ updating the given movie hash.

src/pages/build/querying-source-chains.md
ThetaSinner: Understood that it makes some sense at the moment. The struggle is trying to make Record serve both purposes of being an internal data structure and making it serialize nicely for hApp use. Along with being cumbersome as you say. I'll try to leave this one alone for now :)

src/pages/build/querying-source-chains.md
pdaoust: Always good to have extra details 👍🏼

src/pages/build/querying-source-chains.md
pdaoust: Can you suggest a legit but simple use case? The only ones I can think of involve transactions, which are a rabbit hole in the extreme.

src/pages/build/querying-source-chains.md
pdaoust: https://github.com/holochain/docs-pages/pull/535/commits/d7d65893493951711fb19feb3cf9302244db1736

src/pages/build/querying-source-chains.md
pdaoust: removed in https://github.com/holochain/docs-pages/pull/535/commits/f71509a52d0b9a6c7519e7ef6f62557f476a95f8

src/pages/build/querying-source-chains.md
pdaoust: removed in https://github.com/holochain/docs-pages/pull/535/commits/f71509a52d0b9a6c7519e7ef6f62557f476a95f8

src/pages/build/querying-source-chains.md
pdaoust: Removed in https://github.com/holochain/docs-pages/pull/535/commits/f71509a52d0b9a6c7519e7ef6f62557f476a95f8 and I've created an issue #537 so we don't miss writing about it in the future.

pdaoust: removed in https://github.com/holochain/docs-pages/pull/535/commits/f71509a52d0b9a6c7519e7ef6f62557f476a95f8 ; created an issue for documenting `get_agent_activity`

ThetaSinner (approved):
🥳

PR #539, opened 2025-02-28
Fix: make 'connecting the parts' title consistent

pdaoust:
Just correcting something that didn't get changed everywhere.


src/pages/build/index.md
```diff
@@ -42,7 +42,7 @@ Now that you've got some basic concepts and the terms we use for them, it's time
 * [Links, Paths, and Anchors](/build/links-paths-and-anchors/) --- creating relationships between data
 :::
 
-## Connecting everything together
+## Connecting the parts
```
c12i, line 45: What do you think of this one?

```suggestion
## Connecting the dots
```

c12i (approved):

matthme (approved):

src/pages/build/index.md
pdaoust: It's fun; I'm not sure if it clarifies the intent of this category though. That said, I'm not sure that either 'Connecting everything together' (which was too long and making the nav menu look ugly) or 'Connecting the parts' do either. This PR is just to fix a naming discrepancy, but I'm totally willing to come up with a new name if a succinct and precise one comes along. The topic of this section is interfacing with a zome in a cell -- both calls (which includes capabilities) and signals. Any ideas?

c12i: I see, no problem keeping it as is then. `Connecting the parts` communicates other parts covered beforehand.

PR #540, opened 2025-03-05
Build Guide: Cloning

### LINKED ISSUE #525: Build guide: working with clones
* DNA modifiers (not what they're for; that's covered in the DNAs page)
* Creating a clone
* Disabling/enabling a clone
* Deleting a clone and what happens to the data
* Give examples for both the front end and the HDK

pdaoust:
Closes #525 .

### DIFF
--- a/src/pages/build/cloning.md
+++ b/src/pages/build/cloning.md
@@ -1,0 +1,239 @@
+---
+title: "Cloning"
+---
+
+::: intro
+**Cloning** an existing [DNA](/build/dnas) creates a new DNA with its own independent network and instantiates a new cell from this DNA. You can optionally change the [DNA properties](/build/dnas/#use-dna-properties) when you clone.
+:::
+
+## Creating independent networks with the same rules
+
+As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and [DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same  but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
+
+An agent creates a clone by choosing an existing DNA in a hApp, then specifying at least one new [DNA modifier](/build/dnas/#integrity-section). Then they have to share the modifier(s) with all other peers who want to join the network, and those peers specify the exact same modifiers in order to create an identical clone with an identical DNA hash.
+
+!!! info Creating vs joining a clone network
+From the perspective of using the functions we talk about in the rest of this page, there's really no meaningful difference between _creating_ a network and _joining_ it. Agents 'become the network' together, simply by instantiating a cell from a DNA, discovering other agent cells with the same DNA hash, and beginning to communicate with each other.
+!!!
+
+Here are the three DNA modifiers and how to use them:
+
+### Network seed
+
+When all you want to do is create a new network space, simply specify a new **network seed**. It's an arbitrary vector of bytes that serves no purpose but to change the DNA hash. You can think of it like a passcode to the network.
+
+### DNA properties
+
+The DNA properties are constants that your integrity and coordinator zomes [can access](/build/dnas/#use-dna-properties) to change their behavior. Because they can be used in validation logic, they affect the 'meaning' of your integrity code, which is why they affect the DNA hash (a network is defined by its validation logic).
+
+You can specify DNA properties without specifying a network seed, but be aware that this doesn't guarantee a unique network space. If you're cloning in order to both modify DNA behavior _and_ create a 'private' network space, specify a unique network seed along with the DNA properties.
+
+### Origin time
+
+While you can specify an **origin time** (the earliest valid timestamp for any DHT data), in practice this isn't necessary. Holochain will just use the origin time from the original DNA.
+
+## Clone a DNA from a client
+
+If you want to clone a DNA from the client side, use the [`AppWebsocket.prototype.createCloneCell`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.createclonecell.md).
+
+This example shows two functions: one that creates or joins a private chat room from a DNA whose role in the hApp manifest is named `chat`. It uses the `getHolochainClient` helper we [created in the Front End page](/build/connecting-a-front-end/#connect-to-a-happ-with-the-javascript-client).
+
+```typescript
+
+async function createOrJoinPrivateChatRoom(
+    // A human-readable name. We can use this in the UI.
+    // When we create the clone, we pass it to Holochain.
+    // It's local to the agent and doesn't affect anything functional, so
+    // different agents can give the same chat room clone a different name
+    // and still be able to access the room.
+    name: String,
+    network_seed?: NetworkSeed
+): Promise<ClonedCell> {
+    if (typeof network_seed == "undefined") {
+        // The person who creates a chat room should come up with a unique
+        // network seed. This won't prevent unauthorized access if existing
+        // members share it with others, but it will prevent accidental
+        // clashes.
+        network_seed = crypto.randomUUID();
+    }
+
+    let client = await getHolochainClient();
+    return await client.createCloneCell({
+        modifiers: { network_seed, },
+        name,
+        role_name: "chat",
+    });
+}
+```
+
+## Clone a DNA from a coordinator zome
+
+You can also clone a DNA from within your back end with the [`create_clone_cell`](https://docs.rs/hdk/latest/hdk/clone/fn.create_clone_cell.html) host function.
+
+```rust
+use hdk::prelude::*;
+
+#[derive(Serialize, Deserialize)]
+pub struct CreateOrJoinPrivateChatRoomInput {
+    pub name: String,
+    pub network_seed: Option<NetworkSeed>,
+    pub chat_dna_hash: DnaHash,
+}
+
+#[hdk_extern]
+pub fn create_or_join_private_chat_room(input: CreateOrJoinPrivateChatRoomInput) -> ExternResult<ClonedCell> {
+    let network_seed = input.network_seed
+        .map(|s| s.as_bytes())
+        .unwrap_or(random_bytes()?.into());
+
+    let cell_id = CellId::new(input.chat_dna_hash, agent_info()?.agent_latest_pubkey);
+    let create_clone_cell_input = CreateCloneCellInput {
+        cell_id: cell_id,
+        membrane_proof: None,
+        name: input.name.into(),
+    };
+    create_clone_cell(create_clone_cell_input)
+}
+```
+
+!!! info The `create_clone_cell` host function requires the DNA hash and agent ID
+Unlike the JS client's `createCloneCell` API method, the HDK's `create_clone_cell` host function can't refer to an existing DNA by its role name. Instead, you have to specify the DNA hash and agent ID of an existing cell. <!-- TODO: change this if holochain/holochain#4762 gets addressed --> Because it's impossible to get the DNA hash of another role in the hApp using the HDK, you need to either hard-code it in the coordinator zome (which leads to [tight coupling](https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29#Disadvantages_of_tight_coupling)) or pass it in from the client. Here's some sample code to get a DNA hash from a role name and pass it to the function we just defined:
+
+```typescript
+async function createOrJoinPrivateChatRoom(name: String, network_seed: NetworkSeed?): Promise<ClonedCell> {
+    let client = await getHolochainClient();
+    let appInfo = await client.appInfo();
+    // The first cell filling a role will have the 'prototype' DNA.
+    let chat_cell = appInfo.cell_info["chat"]?[0]?;
+    if (typeof chat_cell == 'undefined') {
+        throw new Error("Couldn't find a cell with the role name 'chat'.");
+    }
+    let chat_dna_hash = (chat_cell.type == CellType.Provisioned || chat_cell.type == CellType.Cloned)
+        ? chat_cell.value.cell_id[0]
+        : chat_cell.value.dna;
+    // This is the zome function we defined in the last code block.
+    // Let's assume it's in a DNA whose role name is `chat_lobby`.
+    return await client.callZome(
+        role_name: "chat_lobby",
+        zome_name: "chat_lobby",
+        fn_name: "create_or_join_private_chat_room",
+        payload: {
+            name,
+            network_seed,
+            chat_dna_hash,
+        }
+    );
+}
+```
+!!!
+
+## Get all the clones of a role
+
+To get all the clones of a given DNA, use [`AppWebsocket.prototype.appInfo`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appclient.appinfo.md) and take a look at the return value's `cell_info` property, which is an object that maps roles to the cells belonging to those roles.
+
+```typescript
+async function getPrivateChatCells(): Promise<Array<ClonedCell>> {
+    let client = await getHolochainClient();
+    let appInfo = await client.appInfo();
+    let chats = appInfo.cell_info["chat"]?;
+    if (typeof chats == "undefined") {
+        throw new Error("The chat role isn't defined in this hApp.");
+    }
+
+    return chats
+        .filter((cell) => cell.type == CellType.Cloned)
+        .map((cell) => cell.value);
+}
+```
+
+## Address a clone cell when calling a zome function
+
+To use a clone cell, you can address it either by its new DNA hash or its **clone ID**, which is a concatenation of its role name and a **clone index**. This value is _local to the agent's hApp instance_ and is returned by the conductor when the cell has been cloned.
+
+### In the client {#call-clone-from-client}
+
+This example posts a message to a given chat. It assumes that the DNA with the `chat` role has a `chat` zome with a function called `post_message` that accepts a string and returns the message hash.
+
+```typescript
+async function postMessage(chatId: CellId | RoleName, message: String) -> Promise<ActionHash> {
+    let client = await getHolochainClient();
+    // Check whether we've been passed a role name / clone ID (which is a
+    // string) or a cell ID (which is an array). Use this to construct the
+    // right payload for callZome.
+    let chatIdProperty = typeof chatID === "string"
+        ? { role_name: chatId, }
+        : { cell_id: chatId, };
+    let callZomePayload = {
+        ...chatIdProperty,
+        zome_name: "chat",
+        fn_name: "post_message",
+        payload: message,
+    }
+    return await client.callZome(callZomePayload);
+}
+```
+<!-- TODO: update this if holochain/holochain#4772 is accepted and done -->
+
+### In a coordinator zome {#call-clone-from-coordinator}
+
+The HDK's [`call`](https://docs.rs/hdk/latest/hdk/p2p/fn.call.html) host function takes roughly the same arguments. This example sends a regular status message to a given chat, perhaps triggered by a client that runs on a schedule.
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn status_message(chatId: CallTargetCell) -> ExternResult<ActionHash> {
+    let message = format!("The time is {} and I'm still running", sys_time()?);
+    call(
+        to_cell: chatId,
+        zome_name: "chat".into(),
+        fn_name: "post_message".into(),
+        cap_secret: None,
+        payload: message,
+    )
+}
+```
+
+## Disable a clone cell
+
+When an agent no longer wants to be part of a network, you can disable the clone. The cell remains in the database but stops responding to zome calls and peer-to-peer network traffic. The hApp can't delete clone cells (to prevent malicious or malformed zomes or front ends from deleting data), but a Holochain runtime will clean up data when a hApp is uninstalled.
+
+### In the client {#disable-clone-from-client}
+
+Use the [`AppWebsocket.prototype.disableCloneCell`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.disableclonecell.md) method to disable a cell from a front end.
+
+```typescript
+async function pauseChat(chatId: CloneCellId): Promise<void> {
+    let client = await getHolochainClient();
+    return client.disableCloneCell({ clone_cell_id: chatId });
+}
+```
+
+### In a coordinator zome {#disable-clone-from-coordinator}
+
+Use the [`disable_clone_cell`](https://docs.rs/hdk/latest/hdk/clone/fn.disable_clone_cell.html) host function to disable a cell from a coordinator zome within the same hApp.
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn pause_chat(chat_id: CloneCellId) -> ExternResult<()> {
+    let input = DisableCloneCellInput { clone_cell_id: chat_id };
+    disable_clone_cell(input)
+}
+```
+
+## Reference
+
+* [`AppWebsocket.prototype.createCloneCell`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.createclonecell.md)
+* [`hdk::clone::create_clone_cell`](https://docs.rs/hdk/latest/hdk/clone/fn.create_clone_cell.html)
+* [`AppWebsocket.prototype.appInfo`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appclient.appinfo.md)
+* [`AppWebsocket.prototype.callZome`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.callzome.md)
+* [`hdk::p2p::call`](https://docs.rs/hdk/latest/hdk/p2p/fn.call.html)
+* [`AppWebsocket.prototype.disableCloneCell`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appwebsocket.disableclonecell.md)
+* [`hdk::clone::disable_clone_cell`](https://docs.rs/hdk/latest/hdk/clone/fn.disable_clone_cell.html)
+
+## Further reading
+
+* [Build Guide: hApps: `clone_limit`](/build/happs/#clone-limit)
+* [Build Guide: DNAs: Single vs multiple DNAs](/build/dnas/#single-vs-multiple-dnas)
--- a/src/pages/build/dnas.md
+++ b/src/pages/build/dnas.md
@@ -34,7 +34,7 @@ It'll also add the new DNA to `workdir/happ.yaml`.
 
 ## Specify a DNA manifest
 
-A DNA manifest is written in [YAML](https://yaml.org/). It contains metadata about the DNA, a section for **integrity modifiers**, and a list of coordinator zomes for interacting with the DNA.
+A DNA manifest is written in [YAML](https://yaml.org/). It contains metadata about the DNA, a section for **integrity** code and modifiers, and a list of [**coordinator zomes**](/build/zomes/#coordinator).
 
 If you want to write your own manifest file, name it `dna.yaml` and give it the following structure. This example assumes that all of your zomes are in a folder called `zomes/`. Afterwards we'll explain what the fields mean.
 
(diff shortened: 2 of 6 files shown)


src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+title: "Cloning"
+---
+
+::: intro
+**Cloning** an existing [DNA](/build/dnas) creates a new DNA with its own independent network and instantiates a new cell from this DNA. You can optionally change the [DNA properties](/build/dnas/#use-dna-properties) when you clone.
+:::
+
+## Creating independent networks with the same rules
+
+As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and [DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same  but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
```
mattyg, line 11:
```suggestion
As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same  but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
```

Not sure if this was supposed to link to something or typo

mattyg, line 11:
```suggestion
As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and [DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+!!! info Creating vs joining a clone network
+From the perspective of using the functions we talk about in the rest of this page, there's really no meaningful difference between _creating_ a network and _joining_ it. Agents 'become the network' together, simply by instantiating a cell from a DNA, discovering other agent cells with the same DNA hash, and beginning to communicate with each other.
+!!!
+
+Here are the three DNA modifiers and how to use them:
+
+### Network seed
+
+When all you want to do is create a new network space, simply specify a new **network seed**. It's an arbitrary vector of bytes that serves no purpose but to change the DNA hash. You can think of it like a passcode to the network.
```
mattyg, line 23:
```suggestion
When all you want to do is create a new network space, simply specify a new **network seed**. It's an arbitrary string that serves no purpose but to change the DNA hash. You can think of it like a passcode to the network.
```

It is a bit odd, but its a string, not a byte array

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+### DNA properties
+
+The DNA properties are constants that your integrity and coordinator zomes [can access](/build/dnas/#use-dna-properties) to change their behavior. Because they can be used in validation logic, they affect the 'meaning' of your integrity code, which is why they affect the DNA hash (a network is defined by its validation logic).
+
+You can specify DNA properties without specifying a network seed, but be aware that this doesn't guarantee a unique network space. If you're cloning in order to both modify DNA behavior _and_ create a 'private' network space, specify a unique network seed along with the DNA properties.
```
mattyg, line 29:
```suggestion
```
I find this more confusing than clarifying. I think the network seed section above explains it enough.

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+    // When we create the clone, we pass it to Holochain.
+    // It's local to the agent and doesn't affect anything functional, so
+    // different agents can give the same chat room clone a different name
+    // and still be able to access the room.
+    name: string,
+    network_seed?: NetworkSeed
+): Promise<ClonedCell> {
+    if (typeof network_seed == "undefined") {
+        // The person who creates a chat room should come up with a unique
+        // network seed. This won't prevent unauthorized access if existing
```
mattyg, line 60:
```suggestion
        // network seed. This ensures it will be a separte network, independent from all other chat rooms.
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+        // members share it with others, but it will prevent accidental
```
mattyg, line 61:
```suggestion
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+        // clashes.
```
mattyg, line 62:
```suggestion
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+        network_seed = crypto.randomUUID();
+    }
+
+    let client = await getHolochainClient();
+    return await client.createCloneCell({
+        modifiers: { network_seed, },
```
mattyg, line 68:
```suggestion
        modifiers: { network_seed },
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+## Clone a DNA from a coordinator zome
+
+You can also clone a DNA from within your back end with the [`create_clone_cell`](https://docs.rs/hdk/latest/hdk/clone/fn.create_clone_cell.html) host function. You need to enable the `properties` feature in your zome's `Cargo.toml` file:
+
+```toml
+[dependencies]
+hdk = { workspace = true, features = ["properties"] }
```
mattyg, line 81: Wondering if this should be a github issue? I feel like this should be a default feature.

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+    pub chat_dna_hash: DnaHash,
+}
+
+#[hdk_extern]
+pub fn create_or_join_private_chat_room(input: CreateOrJoinPrivateChatRoomInput) -> ExternResult<ClonedCell> {
+    let network_seed_bytes = input.network_seed
+        .map(|s| s.as_bytes().to_vec())
+        .unwrap_or(random_bytes(32)?.into_vec());
+
+    let network_seed = std::str::from_utf8(&network_seed_bytes)
```
mattyg, line 102: I would use the `uuid` crate to simplify this example: https://docs.rs/uuid/latest/uuid/

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+#[hdk_extern]
+pub fn pause_chat_by_dna_hash(dna_hash: DnaHash) -> ExternResult<()> {
+    let input = DisableCloneCellInput {
+        clone_cell_id: CloneCellId::DnaHash((dna_hash)),
+    };
+    disable_clone_cell(input)
+}
+```
+
```
mattyg, line 300: Missing enableCloneCell?

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+async function createOrJoinPrivateChatRoom(
+    // A human-readable name. We can use this in the UI.
+    // When we create the clone, we pass it to Holochain.
+    // It's local to the agent and doesn't affect anything functional, so
+    // different agents can give the same chat room clone a different name
+    // and still be able to access the room.
+    name: string,
+    network_seed?: NetworkSeed
+): Promise<ClonedCell> {
+    if (typeof network_seed == "undefined") {
```
mattyg, line 58:
```suggestion
    if (network_seed === undefined) {
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+    };
+    create_clone_cell(create_clone_cell_input)
+}
+```
+
+!!! info The `create_clone_cell` host function requires the DNA hash and agent ID
+Unlike the JS client's `createCloneCell` API method, the HDK's `create_clone_cell` host function can't refer to an existing DNA by its role name. Instead, you have to specify the DNA hash and agent ID of an existing cell. <!-- TODO: change this if holochain/holochain#4762 gets addressed --> Because it's impossible to get the DNA hash of another role in the hApp using the HDK, you need to either hard-code it in the coordinator zome (which is not recommended because it leads to [tight coupling](https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29#Disadvantages_of_tight_coupling)) or pass it in from the client. Here's a sample client-side function to get a DNA hash from a role name and pass it to the zome function we just defined:
+
+```typescript
+import { CellType } from "@holochain/client";
```
mattyg, line 123:
```suggestion
import { CellType, type NetworkSeed, type ClonedCell } from "@holochain/client";
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+!!!
+
+## Get all the clones of a role
+
+To get all the clones of a given DNA, use [`AppWebsocket.prototype.appInfo`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appclient.appinfo.md) and take a look at the return value's `cell_info` property, which is an object that maps roles to the cells belonging to those roles.
+
+```typescript
+import { ClonedCell } from "@holochain/client";
```
mattyg, line 162:
```suggestion
import { CellType, type ClonedCell } from "@holochain/client";
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+## Address a clone cell when calling a zome function
+
+To use a clone cell, you can address it either by its new DNA hash or its **clone ID**, which is a concatenation of its role name and a **clone index**. This value is _local to the agent's hApp instance_ and is returned by the conductor when the cell has been cloned.
+
+### In the client {#call-clone-from-client}
+
+This example posts a message to a given chat. It assumes that any DNAs cloned from the `chat` role have a `chat` zome with a function called `post_message` that accepts a string and returns the message hash.
+
+```typescript
+import { ActionHash } from "@holochain/client";
```
mattyg, line 187:
```suggestion
import { ActionHash, DnaHash } from "@holochain/client";
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+    let input = DisableCloneCellInput {
+        clone_cell_id: CloneCellId::CloneId(CloneId(clone_role_name)),
+    };
+    disable_clone_cell(input)
+}
+
+#[hdk_extern]
+pub fn pause_chat_by_dna_hash(dna_hash: DnaHash) -> ExternResult<()> {
+    let input = DisableCloneCellInput {
+        clone_cell_id: CloneCellId::DnaHash((dna_hash)),
```
mattyg, line 295:
```suggestion
        clone_cell_id: CloneCellId::DnaHash(dna_hash),
```

mattyg (changes requested):
couple comments, but lookin' good!

src/pages/build/cloning.md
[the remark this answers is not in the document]
matthme: Well, if it's a happ dev focused guide I could personally live with leaving this away for now because the admin websocket introduces an additional layer of complexity that may end up being more confusing than helpful for happ devs maybe?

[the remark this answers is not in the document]
matthme: Agreed, I think this would be useful to add.

matthme (changes requested):

src/pages/build/cloning.md
ThetaSinner: Objection to "private network space". The network seed does little to provide privacy. If you want to keep your DHT data private it needs to be encrypted. Even membrane proofs won't currently do much.

If you think about how we handle passwords in a centralized database. A salt has to be used to make it hard to reverse engineer the password. With shared inputs like properties and zomes, it is possible to figure out the network seed from the resulting DNA hash. In fact, that isn't even really necessary, knowledge of the DNA hash is enough to read the data. There is no further authentication beyond knowledge of that 32 bit value. We don't document that anywhere as secret and Holochain doesn't promise not to reveal it. E.g. it's posted to the bootstrap server, so if nothing else we could deploy a patched bootstrap server that dumps everyone's DNA hashes. Of course we shouldn't want to do that kind of sabotage but it's worth assuming that somebody might do that to their users.

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+When all you want to do is create a new network space, simply specify a new **network seed**. It's an arbitrary vector of bytes that serves no purpose but to change the DNA hash. You can think of it like a passcode to the network.
+
+### DNA properties
+
+The DNA properties are constants that your integrity and coordinator zomes [can access](/build/dnas/#use-dna-properties) to change their behavior. Because they can be used in validation logic, they affect the 'meaning' of your integrity code, which is why they affect the DNA hash (a network is defined by its validation logic).
+
+You can specify DNA properties without specifying a network seed, but be aware that this doesn't guarantee a unique network space. If you're cloning in order to both modify DNA behavior _and_ create a 'private' network space, specify a unique network seed along with the DNA properties.
+
+### Origin time
```
ThetaSinner, line 31: Just an FYI that this has been removed in K2 and won't be needed from 0.5

ThetaSinner: It is correctly documented here though :)

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+!!! info
+All these examples use the [`createHolochainClient` helper from the Connecting a Front End page](/build/connecting-a-front-end/#connect-to-a-happ-with-the-javascript-client).
+!!!
+
+```typescript
+import type { NetworkSeed, ClonedCell } from "@holochain/client";
+import { AppWebsocket } from "@holochain/client";
+
+async function createOrJoinPrivateChatRoom(
```
ThetaSinner, line 49:
```suggestion
async function createOrJoinChatRoom(
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+!!!
+
+## Get all the clones of a role
+
+To get all the clones of a given DNA, use [`AppWebsocket.prototype.appInfo`](https://github.com/holochain/holochain-client-js/blob/main/docs/client.appclient.appinfo.md) and take a look at the return value's `cell_info` property, which is an object that maps roles to the cells belonging to those roles.
+
+```typescript
+import { ClonedCell } from "@holochain/client";
+
+async function getPrivateChatCells(): Promise<Array<ClonedCell>> {
```
ThetaSinner, line 164:
```suggestion
async function getChatCells(): Promise<Array<ClonedCell>> {
```

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+    match response {
+        ZomeCallResponse::Ok(payload) => payload.decode().map_err(|e| wasm_error!(e.to_string())),
+        _ => Err(wasm_error!("Something unexpected happened: {}", response.to_string())),
+    }
+}
+```
+
+## Disable a clone cell
+
+When an agent no longer wants to be part of a network, they can disable the clone. The cell remains in the database but stops responding to zome calls and peer-to-peer network traffic. The hApp can't delete clone cells (to prevent malicious or malformed zomes or front ends from deleting data), but Holochain will clean up data when a hApp is uninstalled.
```
ThetaSinner, line 250: You can delete clone cells though, it's just an admin operation and so your runtime needs to make it available somehow. It's accurate that this isn't exposed on the app interface. It is through the HDK.

src/pages/build/cloning.md
matthme: > With shared inputs like properties and zomes, it is possible to figure out the network seed from the resulting DNA hash

I actually doubt that if the network seed has enough entropy, like for example a `uuid`

> In fact, that isn't even really necessary, knowledge of the DNA hash is enough to read DHT data

Yes but that's the point I think. How do you figure out the dna hash without knowledge of the network seed?

Of course, as you say that relies on the integrity of the bootstrap server to not expose dna hashes...but calling it not private at all seems not appropriate to me either. There is at the very least a high degree of friction to read content of networks with high-entropy network seeds.

src/pages/build/cloning.md
pdaoust: My big issue with it is that it forces me to bring in `serde_yaml` when all I want to do is specify a network seed. I think the `properties` feature should be on `DnaModifiersOpt::properties`, not on `CreateCloneCellInput::modifiers`.

src/pages/build/cloning.md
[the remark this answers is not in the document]
pdaoust: that's right, I'm expressly avoiding anything that invovles the admin API for this first pass at the guide, because when you bring in the admin API you also have to talk about Kangaroo or p2p Shipyard.

src/pages/build/cloning.md
pdaoust: Hooray! I'll leave a TODO

src/pages/build/cloning.md
pdaoust: Well I'll be, [you're right](https://docs.rs/hdk/latest/hdk/clone/fn.delete_clone_cell.html). That seems... ambiguous.

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
+---
+title: "Cloning"
+---
+
+::: intro
+**Cloning** an existing [DNA](/build/dnas) creates a new DNA with its own independent network and instantiates a new cell from this DNA. You can optionally change the [DNA properties](/build/dnas/#use-dna-properties) when you clone.
```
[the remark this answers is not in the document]
pdaoust: I'm inclined to agree about not bothering to talk about cloning from the HDK @matthme -- I'm not sure what the use case is, and without the `properties` feature flag it's essentially useless because other peers can't input the same modifiers and get the same DNA -- so it's valuable for creating a clone but not for joining one.

Here was the original pull: https://github.com/holochain/holochain/issues/3095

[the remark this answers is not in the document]
pdaoust: Re: language, I'm sitting with this... for me the most natural and correct way to think about it is to clone a DNA, not a cell, and create a new clone. I know that at this point the biological analogy disintegrates into a steaming pile of compost, but in the guts of Holochain that's what's happening -- the DNA 'exists' in the sense of being a known template for creating cells, and you can clone (really more like 'mutate') that DNA and use it as a template for creating new sells of a slightly different nature.

Does that language work for you @matthme ? Should I expand it like this, or does it still not work and I should just work on the suggestion you gave?

src/pages/build/cloning.md
pdaoust: Would that work inside a zome? I'm wondering where `uuid` gets its entropy from.

mattyg: i think so: https://docs.rs/uuid/latest/uuid/#webassembly

src/pages/build/cloning.md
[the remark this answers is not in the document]
mattyg: In that case, I would at least mention that you cannot delete clones from the app api.

src/pages/build/cloning.md
pdaoust: ah yes, I did see that, and sadly it looks like it [uses the host's `WebCrypto` API via wasm-bindgen](https://github.com/uuid-rs/uuid/blob/c36beb14d50f835c1f1220117ca51aae64860a3e/src/rng.rs#L174-L318).

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+        fn_name: "create_or_join_private_chat_room",
+        payload: {
+            name,
+            network_seed,
+            chat_dna_hash,
+        }
+    });
+}
+```
+!!!
```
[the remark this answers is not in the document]
pdaoust: No, this is intentional -- the code block is part of the aside. I'd like a better way of displaying in-depth info -- or rather, _not_ displaying it until the reader wants to go deep. This really should be hidden in an accordion.

src/pages/build/cloning.md
pdaoust: typo; thanks for catching it. Fixed in https://github.com/holochain/docs-pages/pull/540/commits/79a00512074d3d5b208f271579a65f4c58051d13

pdaoust: There was supposed to be the word 'code' there; fixed as part of https://github.com/holochain/docs-pages/pull/540/commits/f81352c3d05b877bcc9cfcac06d5ddb83aa36a06

src/pages/build/cloning.md
pdaoust: @matthme

> I think that changing the dna properties does actually guarantee a unique network space but that's not it's primary purpose and so what I hear you pointing to @pdaoust is that one should be aware of the latter in order to not accidentally collide with the network of cells cloned on other peers conductors.

Exactly right. It's unique to itself but will be identical to any DNA cloned with the exact same properties, and that isn't desirable if you expected a private space for your group. I'll go back and work on the text and see what I come up with.

@ThetaSinner I hear what you're saying about 'private', but I agree with @matthme that privacy is a gradient, not a binary. I've removed the word `private` from the function names, because really this pattern works for public clones too. I'd still like to mention something about network seeds and privacy, because a seed with sufficient entropy combined with a conductor and bootstrap server that don't leak data _are_ somewhere along the privacy gradient -- but I'll put it an aside.

Removed 'private' stuff in https://github.com/holochain/docs-pages/pull/540/commits/b318995bc2f01825d35b08b3bc53a9ed437fb78e

src/pages/build/cloning.md
pdaoust: Given that this is outside the scope of the documentation, can I resolve this thread?

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+:::
+
+## Creating independent networks with the same rules
+
+As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and [DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same  but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
+
+An agent creates a clone by choosing an existing DNA in a hApp, then specifying at least one new [DNA modifier](/build/dnas/#integrity-section). Then they have to share the modifier(s) with all other peers who want to join the network, and those peers specify the exact same modifiers in order to create an identical clone with an identical DNA hash.
+
+!!! info Creating vs joining a clone network
+From the perspective of using the functions we talk about in the rest of this page, there's really no meaningful difference between _creating_ a network and _joining_ it. Agents 'become the network' together, simply by instantiating a cell from a DNA, discovering other agent cells with the same DNA hash, and beginning to communicate with each other.
```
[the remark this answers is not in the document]
pdaoust: Thanks! 🙏🏼

src/pages/build/cloning.md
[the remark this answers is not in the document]
pdaoust: Mentioned in https://github.com/holochain/docs-pages/pull/540/commits/98b2bccd98d6ac23b04e75e4989bfdd2d077c94d#diff-da16bdefaab41ed88cbb7f81466da2d66c78c14a28afb04ea25a4b60ec38e559R199 ; I've also added a `delete_clone_cell` example in https://github.com/holochain/docs-pages/pull/540/commits/ada9dff40f6da401a2462001b7dd12e9b72c79c1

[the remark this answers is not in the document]
pdaoust: added in https://github.com/holochain/docs-pages/pull/540/commits/4930aa74ebf559216587e93d001c7e5fb91700f2

src/pages/build/cloning.md
pdaoust: documented in https://github.com/holochain/docs-pages/pull/540/commits/ada9dff40f6da401a2462001b7dd12e9b72c79c1

src/pages/build/cloning.md
pdaoust: todo added https://github.com/holochain/docs-pages/pull/540/commits/78149c5d416d9e9a0f85acedaefd0ca383573bc0

src/pages/build/cloning.md
pdaoust: More work on the implication of 'private' DHTs -- toning down/eliminating claims but not removing the idea that it has some value for privacy. https://github.com/holochain/docs-pages/pull/540/commits/39919877028f37f0951fb2b71e5a3f918d333186

src/pages/build/cloning.md
pdaoust: Added in https://github.com/holochain/docs-pages/pull/540/commits/0c9b4d3b24d7cb9914e0360d9db7173503aabddf

src/pages/build/cloning.md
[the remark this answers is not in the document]
matthme: I would say that when I read your comment now I understand what you mean and it seems reasonable to me. I just did not interpret it that way intuitively...so I think I'm happy with it staying as it is.

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+
+::: intro
+**Cloning** an existing [DNA](/build/dnas) creates a new DNA with its own independent network and instantiates a new cell from this DNA. You can optionally change the [DNA properties](/build/dnas/#use-dna-properties) when you clone.
+:::
+
+## Creating independent networks with the same rules
+
+As we described in [DNAs](/build/dnas/), every DNA has an independent network, keyed by the hash of all the [integrity zomes](/build/zomes/#integrity) and [DNA modifiers. Cloning allows you to change an existing DNA slightly, so that it executes the same  but enjoys a separate membership of peers and [DHT database](/concepts/4_dht/).
+
+An agent creates a clone by choosing an existing DNA in a hApp, then specifying at least one new [DNA modifier](/build/dnas/#integrity-section). Then they have to share the modifier(s) with all other peers who want to join the network, and those peers specify the exact same modifiers in order to create an identical clone with an identical DNA hash.
```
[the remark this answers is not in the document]
pdaoust: attempted to fix in https://github.com/holochain/docs-pages/pull/540/commits/ed46b34d8acef613d36ce47986a37da140d72586

src/pages/build/cloning.md
```diff
@@ -0,0 +1,314 @@
...
+        modifiers,
+        membrane_proof: None,
+        name: input.name.into(),
+    };
+    create_clone_cell(create_clone_cell_input)
+}
+```
+
+!!! info The `create_clone_cell` host function requires the DNA hash and agent ID
+Unlike the JS client's `createCloneCell` API method, the HDK's `create_clone_cell` host function can't refer to an existing DNA by its role name. Instead, you have to specify the DNA hash and agent ID of an existing cell. <!-- TODO: change this if holochain/holochain#4762 gets addressed --> Because it's impossible to get the DNA hash of another role in the hApp using the HDK, you need to either hard-code it in the coordinator zome (which is not recommended because it leads to [tight coupling](https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29#Disadvantages_of_tight_coupling)) or pass it in from the client. Here's a sample client-side function to get a DNA hash from a role name and pass it to the zome function we just defined:
```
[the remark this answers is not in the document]
pdaoust: Hopefully https://github.com/holochain/docs-pages/pull/540/commits/ed46b34d8acef613d36ce47986a37da140d72586 fixes this?

matthme (approved):

src/pages/build/cloning.md
mattyg: Doesn't look like this was resolved? @pdaoust

src/pages/build/dnas.md
```diff
@@ -134,14 +134,20 @@ So Holochain manages the dependency mapping for you, allowing you to write code
...
 * **Creating privileged spaces.** A chat hApp may have both public and private rooms, all [cloned](/resources/glossary/#cloning) from a single `chat_room` DNA. This is a special case, as they all use just one base DNA, but they change just one [integrity modifier](#dna-manifest-structure-at-a-glance) such as the network seed to create new DNAs.
 * **Discarding or archiving data.** Because no data is ever deleted in a cell or the network it belongs to, a lot of old data can accumulate. Creating clones of a single storage-heavy DNA, bounded by topic or time period, allows agents to participate in only the networks that contain the information they need. As agents leave networks, unwanted data naturally disappears from their machines.
 
+!!! Using the network seed for private spaces
+A network seed with high entropy could be used as a passcode for joining a network, allowing you to create a moderately private space. It shouldn't be considered a truly secret space though, as data isn't encrypted at rest and anyone with access to the network seed or the resulting modified DNA hash will be able to join the network and access its data and membership.
```
mattyg, line 146:
```suggestion
A network seed with high entropy could be used as a passcode for joining a network, allowing you to create a moderately private space. It shouldn't be considered a truly secret space though, as anyone with access to the network seed or the resulting modified DNA hash will be able to join the network and access its data and membership.
```

Data *is* encrypted at rest in the conductor dbs

pdaoust: 🤦🏼‍♂️ you're right, I forgot about that

src/pages/build/cloning.md
mattyg: I'm happy with this -- just want to make sure @matthme and @ThetaSinner are too

pdaoust: yup, they've given tacit approval to the PR. Thanks for closing this up @mattyg!

mattyg (approved):

PR #542, opened 2025-03-13
fix: un-squodge two sections on the build guide TOC

pdaoust:
It's literally four characters, one of which is a newline.


ThetaSinner (approved):

jost-s (approved):
image
Approved

PR #546, opened 2025-03-21
Host API overview page

pdaoust:
I needed a placeholder for misc functions that didn't fit elsewhere. This is it. Should be a small review. This PR needs to be merged before I address #477 , #538 , and #541 , which are next up in my queue.

PR #547, opened 2025-03-25
Cryptography functions build guide

### LINKED ISSUE #477: Practical guide for using cryptography utilities
Possible topics include:
- Crypto
   - hashing
   - Ed25519 signing/verifying
   - box encryption

pdaoust:
Closes #477 . ~~Depends on #546 and shouldn't be merged until it's merged.~~ (update: #546 is merged now)

Covers:

* Hashing
* Signing
* Encryption


src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+### Hash arbitrary data
+
+You can also Blake2b hash any data you like, with any hash length up to 256 bits, using [`hash_blake2b`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_action.html):
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+fn hash_hello_16_bit() -> ExternResult<Vec<u8>> {
+    hash_blake2b("hello".as_bytes().to_vec(), 16)
```
neonphog, line 99: Pretty sure this returns 16 bytes... not 16 bits

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+There are [other hashing algorithms](https://docs.rs/hdk/latest/hdk/hash/index.html) available:
+
+```rust
+use hdk::prelude::*;
+
+#[derive(Serialize, Deserialize)]
+pub struct ManyHashes {
+    // Used by Ethereum and other EVM blockchains.
+    keccak256: Vec<u8>,
+    sha3_256: Vec<u8>,
```
neonphog, line 112: These should not be in the hdk... *sigh* If you want esoteric stuff you should include it in your wasm.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+
+Holochain keeps all secret material in its own key store so you don't have to keep it safe yourself. We also recommend that, instead of using their agent IDs, all participants generate extra keys specifically for encryption using the [`create_x25519_keypair`](https://docs.rs/hdk/latest/hdk/x_salsa20_poly1305/fn.create_x25519_keypair.html) host function [for better security](https://doc.libsodium.org/quickstart#how-can-i-sign-and-encrypt-using-the-same-key-pair), which corresponds to libsodium's `crypto_box_keypair` function.
+
+!!! info These algorithms are not quantum-resistant
+While these algorithms represent some of the best in terms of problem-free secure encryption, they aren't quantum-resistant, because they use elliptic curve cryptography. We recommend not storing any encrypted data on the DHT if you're concerned about data being harvested and later decrypted with a quantum computer.
+!!!
```
neonphog, line 197: I don't feel like this is necessary... feels FUDy...

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+        encrypted_secret,
+        secret_ref
+    )
+}
+
+// Now that we have an API for establishing a shared secret for chatting, we
+// can start sending messages.
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct
```
neonphog, line 341: ?

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+
+!!! info These algorithms are not quantum-resistant
+While these algorithms represent some of the best in terms of problem-free secure encryption, they aren't quantum-resistant, because they use elliptic curve cryptography. We recommend not storing any encrypted data on the DHT if you're concerned about data being harvested and later decrypted with a quantum computer.
+!!!
+
+In the following examples, we'll show you how to establish secure channels using the two algorithms, but we won't show you how to create groups or send messages --- we'll leave that up to you. (You could use [the chat examples from the Cloning page](/build/cloning/), for example.)
+
+### Sending encrypted messages between two parties using box
```
neonphog, line 201: I'm not super comfortable with this section. It feels like advice.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+    let { my_public_key, recipient_public_key, message } = input;
+    x_25519_x_salsa20_poly1305_decrypt(
+        my_public_key,
+        sender_public_key,
+        message
+    ).into()
+}
+```
+
+### Sending encrypted messages among multiple parties using secretbox
```
neonphog, line 253: I'm very not comfortable with this section. It is not only advice, but could be construed as recommending a pattern. Very much would prefer it be deleted.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+        none,
+        none
+    })
+}
+```
+
+!!! info Why would you hash an entry?
+Usually you have an entry hash and want to retrieve the entry data. When would you have data and want to have it?
+
+Entries don't need to be written to the DHT in order to serve as a [base or target for links](/build/links-paths-and-anchors/#define-a-link-type). As long as the entry can be reconstructed by anyone who wants to store or retrieve links on its basis address, such as the tag in this example, its hash can also be reconstructed and used in `create_link` or `get_links`. This saves some storage and validation overhead for everyone.
```
matthme, line 70: Another common case is that you *do* also want to write it to the DHT but `create_entry()` returns the *action* hash while you want to create a link from the entry hash instead because it's not a thing where it matters "who" created it (or "when") but only "what".

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+        sha2_512: hash_sha512(hello_bytes.clone())?,
+    }
+}
+```
+
+## Sign data
+
+### With an agent key
+
+To sign data with an agent's private key, pass the data and the key to to the [`sign`](https://docs.rs/hdk/latest/hdk/ed25519/fn.sign.html) or [`sign_raw`](https://docs.rs/hdk/latest/hdk/ed25519/fn.sign_raw.html). `sign_raw` signs a `Vec<u8>` while `sign` accepts anything that can be serialized.
```
matthme, line 133:
```suggestion
To sign data with an agent's private key, pass the data and the key to to the [`sign`](https://docs.rs/hdk/latest/hdk/ed25519/fn.sign.html) or [`sign_raw`](https://docs.rs/hdk/latest/hdk/ed25519/fn.sign_raw.html) function. `sign_raw` signs a `Vec<u8>` while `sign` accepts anything that can be serialized.
```

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+use base64::prelude::*;
+use movies_integrity::*;
+
+#[hdk_extern]
+pub fn create_joining_certificate(invitee: AgentPubKey) -> ExternResult<String> {
+    // The `DnaProperties` struct comes from the example we linked to above.
+    let dna_props = DnaProperties::try_from_dna_properties()?;
+    let administrator = dna_props.authorized_joining_certificate_issuer;
+    let my_pub_key = agent_info()?.agent_latest_pubkey;
+    if (administrator != my_pub_key) {
```
matthme, line 148: I think Rust doesn't use brackets, right?

matthme:
```suggestion
    if administrator != my_pub_key {
```

src/pages/build/cryptography-functions.md
matthme: It does not feel too FUDy to me...just transparent. It's also mentioning that this is only a concern if you expect data to get harvested and stored long-term which seems to me like a legitimate concern depending on the use case.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+
+Sending encrypted messages between two parties is straightforward and doesn't require you to generate and share an encryption key --- instead, the two participants generate a new key pair to use specifically for encryption, share the public component with each other, and use [ECDH](https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman) to generate a shared secret non-interactively. Messages are encrypted and decrypted using the [`x_25519_x_salsa20_poly1305_encrypt`](https://docs.rs/hdk/latest/hdk/x_salsa20_poly1305/fn.x_25519_x_salsa20_poly1305_encrypt.html) and [`x_25519_x_salsa20_poly1305_decrypt`](https://docs.rs/hdk/latest/hdk/x_salsa20_poly1305/fn.x_25519_x_salsa20_poly1305_decrypt.html) host functions, which correspond to libsodium's `crypto_box_easy` and `crypto_box_open_easy` functions.
+
+This example implements secure direct messaging using the above functions.
+
+```rust
+use hdk::prelude::*;
+
+#[hdk_extern]
+pub fn create_secure_direct_message_channel() -> ExternResult<XSalsa20Poly1305KeyRef> {
```
matthme, line 211: If the function is calle `create_secure_direct_message_channel()` then it should maybe not just generate the keypair but also store the public part on the source chain or something like that?

matthme: Or create link from a channel anchor to the public key? But then we're starting to get into implying "best practice" for something we may not actually know yet what's best practice?

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Sending encrypted messages between two parties using box
+
+Sending encrypted messages between two parties is straightforward and doesn't require you to generate and share an encryption key --- instead, the two participants generate a new key pair to use specifically for encryption, share the public component with each other, and use [ECDH](https://en.wikipedia.org/wiki/Elliptic-curve_Diffie%E2%80%93Hellman) to generate a shared secret non-interactively. Messages are encrypted and decrypted using the [`x_25519_x_salsa20_poly1305_encrypt`](https://docs.rs/hdk/latest/hdk/x_salsa20_poly1305/fn.x_25519_x_salsa20_poly1305_encrypt.html) and [`x_25519_x_salsa20_poly1305_decrypt`](https://docs.rs/hdk/latest/hdk/x_salsa20_poly1305/fn.x_25519_x_salsa20_poly1305_decrypt.html) host functions, which correspond to libsodium's `crypto_box_easy` and `crypto_box_open_easy` functions.
+
+This example implements secure direct messaging using the above functions.
```
matthme, line 205: Strictly speaking, it's not covering the part of exchanging public keys, right? So it's only partially implementing it.

src/pages/build/cryptography-functions.md
matthme: Which aspect of the "advice" are you not be comfortable with? Using these functions for encrypted direct messaging in general or the specific code pattern as shown below (or something else)?

src/pages/build/cryptography-functions.md
matthme: I think for the purpose of understanding the host functions the section is valuable. Do you think it is actually problematic as a pattern or that there may just be better ones?
Can we maybe add a warning section that the example is only for explanatory purpose and not necessarily to be understood as best practice?

src/pages/build/cryptography-functions.md
neonphog: When you generate a TLS certificate there isn't a big warning telling you that you picked a non-quantum-safe algorithm. Encryption is an advanced feature. If you choose to use it, it is your job to understand the consequences of your algorithm choice.

But I don't need to die on this hill, fine if I'm overruled : )

src/pages/build/cryptography-functions.md
neonphog: This is not too bad, but why does it have to mention channels? Why are there separate pub structs for the inputs? And why are there `#[hkd_extern]`s on there? Is it really a best practice to expose directly the encryption functions as zome calls?

Our liability is limited by us being an open source project, but we still don't want to steer people in the wrong direction. If they implement something badly and have a huge security breach, it's much worse for our reputation if they point to our example code than if they cannot.

If people need examples of API usage, I assert they should be isolated to that exact function with dummy input values.

src/pages/build/cryptography-functions.md
neonphog: See https://github.com/holochain/docs-pages/pull/547/files/89947f2a9fbf3002fdcd9f2ce50a1be679f62481#r2014327313

I vote for deleting the patterns entirely.

src/pages/build/cryptography-functions.md
matthme: Okay, I'm also not having a strong opinion. The difference between TLS and this case here I guess is that TLS is for transport and this is for distributed storage. I guess you could also argue that people may harvest TLS communication and store it permanently for later decryption...but with holochain it's a bit more inherent to the protocol that it may be stored indefinitely somewhere and potentially be decrypted later.

src/pages/build/cryptography-functions.md
pdaoust: @neonphog whoof, you may be right, but that's gonna mean a lot of edits to already approved PRs -- the whole documentation is based on contrived examples like this. I did start out using literals as dummy input values, but the challenges are two: one, they don't compile out of the box, and two, constructing hashes is really noisy looking. (Rust is already noisy enough.) That's why I started creating these contrived zome functions.

I'm not a security expert, but AFAICT the encryption host functions were designed to remove as many foot guns as possible from the underlying libsodium functions that were already encoding some best practices. I do hear you though, recommending something that's poor practice is neither helpful nor legally wise. And as I said, I'm not a security expert :)

Two options here: pare back the examples to use bare host function calls and literals all over the place, or get someone to look over it for security risks.

neonphog: I'm not suggesting you delete all patterns and contrived examples, just the ones here related directly to explicit crypto functions.

I don't believe it is not possible to get them compiling without too much boilerplate. Something like this should work:

```rust
let this_pk = AgentPubKey::from_raw_36(vec![0; 36]);
let peer_pk = AgentPubKey::from_raw_36(vec![1; 36]);
x_25519_x_salsa20_poly1305_encrypt(this_pk, peer_pk, b"hello".to_vec())
```

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
++ holo_hash = { version = "=0.4.0", features = ["hashing"] }
+...
+```
+!!!
+
+### Hash an action or entry
+
+Holochain's native hashing scheme is Blake2b-256.
+
+To hash an entry, use the [`hash_entry`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_entry.html) host function. This example implements a basic tagging taxonomy:
```
mattyg, line 34: IMO we should just have a 3 line example like the one on the docs site. This just adds noise.

```
#[hdk_entry_helper]
struct Foo;

let foo_hash: EntryHash = hash_entry(Foo)?;
```

mattyg: Actually I'm not even sure that example makes sense, as you can just do `let foo_hash: EntryHash = foo.into()`

src/pages/build/cryptography-functions.md
mattyg: Should we raise an issue for removing them @neonphog ?

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+    keccak256: Vec<u8>,
+    sha3_256: Vec<u8>,
+    sha2_256: Vec<u8>,
+    sha2_512: Vec<u8>,
+}
+
+#[hdk_extern]
+fn hash_hello_in_many_ways() -> ExternResult<ManyHashes> {
+    let hello_bytes = "hello".as_bytes().to_vec();
+    ManyHashes {
```
mattyg, line 120: Should be wrapped in Ok()

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct EncryptDirectMessageInput {
+    pub my_public_key: X25519PubKey,
+    pub recipient_public_key: X25519PubKey,
+    pub message: Vec<u8>,
+}
+
+#[hdk_extern]
+pub fn encrypt_direct_message(input: EncryptDirectMessageInput) -> ExternResult<XSalsa20Poly1305EncryptedData> {
+    let { my_public_key, recipient_public_key, message } = input;
```
mattyg, line 227:
```suggestion
    let EncryptDirectMessageInput { my_public_key, recipient_public_key, message } = input;
```

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DecryptDirectMessageInput {
+    pub my_public_key: X25519PubKey,
+    pub sender_public_key: X25519PubKey,
+    pub message: XSalsa20Poly1305EncryptedData,
+}
+
+#[hdk_extern]
+pub fn decrypt_direct_message(input: DecryptDirectMessageInput) -> ExternResult<Vec<u8>> {
+    let { my_public_key, recipient_public_key, message } = input;
```
mattyg, line 244:
```suggestion
    let EncryptDirectMessageInput { my_public_key, recipient_public_key, message } = input;
```

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+    x_25519_x_salsa20_poly1305_decrypt(
+        my_public_key,
+        sender_public_key,
+        message
+    ).into()
```
mattyg, line 249:
```suggestion
    )?
    .ok_or(wasm_error!("Authentication failed, unable to decrypt"))
```
This returns a `Result<Option<_>>`. Also Into<Vec<u8> is not implemented and the inner bytes are not public. Probably should make an issue for that.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct DecryptDirectMessageInput {
+    pub my_public_key: X25519PubKey,
+    pub sender_public_key: X25519PubKey,
+    pub message: XSalsa20Poly1305EncryptedData,
+}
+
+#[hdk_extern]
+pub fn decrypt_direct_message(input: DecryptDirectMessageInput) -> ExternResult<Vec<u8>> {
```
mattyg, line 243:
```suggestion
pub fn decrypt_direct_message(input: DecryptDirectMessageInput) -> ExternResult<XSalsa20Poly1305Data> {
```
Into<Vec> is not implemented and the inner bytes are not public. Probably should make an issue for that.

src/pages/build/cryptography-functions.md
mattyg: I agree this is too much for this page. A simple demonstration of the function call is perfect. We can revisit including something like this when we add "patterns" later.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+}
+```
+
+### Verify a signature
+
+If you have the public key, you can verify a signature against its payload with [`verify_signature`](https://docs.rs/hdk/latest/hdk/ed25519/fn.verify_signature.html) or [`verify_signature_raw`](https://docs.rs/hdk/latest/hdk/ed25519/fn.verify_signature_raw.html). Take a look at the previously mentioned [example from the `genesis_self_check` callback page](/build/genesis-self-check-callback/#joining-certificate) to see it in action.
+
+## Encrypt data
+
+You can encrypt and decrypt data with the [box](https://doc.libsodium.org/public-key_cryptography/sealed_box) and [secretbox](https://doc.libsodium.org/public-key_cryptography/authenticated_encryption) algorithms from [libsodium](https://doc.libsodium.org/). We've selected these because they're robust, well-tested, best-practice algorithms that are fairly easy to use properly. This saves you having to implement your own cryptography scheme.
```
mattyg, line 189:
```suggestion
You can encrypt and decrypt data with the [box](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) and [secretbox](https://doc.libsodium.org/public-key_cryptography/authenticated_encryption) algorithms from [libsodium](https://doc.libsodium.org/). We've selected these because they're robust, well-tested, best-practice algorithms that are fairly easy to use properly. This saves you having to implement your own cryptography scheme.
```

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,409 @@
...
+!!! info Where are the nonces?
+If you're familiar with the box and secretbox algorithms, you'll know that good, cryptographically secure nonces are important, yet they don't appear anywhere in the code above. That's because Holochain generates a nonce and prepends it to the encrypted payload, then extracts the nonce from the payload when it's being decrypted. This ensures that the encryption isn't weakened by insecure nonce implementations.
+!!!
+
+## Reference
+
+* Hashing
+    * [`hdk::hash::hash_entry`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_entry.html)
+    * [`hdk::hash::hash_action`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_action.html)
+    * [`hdk::hash::hash_blake2b`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_action.html)
```
mattyg, line 383:
```suggestion
    * [`hdk::hash::hash_blake2b`](https://docs.rs/hdk/latest/hdk/hash/fn.hash_blake2b.html)
```

src/pages/build/cryptography-functions.md
pdaoust: Only if you have the `hashing` feature flag enabled on the `holo_hash` dependency, which brings in `blake2`. It's not enabled by default.

src/pages/build/cryptography-functions.md
pdaoust: oh nice trick, thank you!

src/pages/build/cryptography-functions.md
pdaoust: Dismissed cuz I got rid of the wrapping function.

src/pages/build/cryptography-functions.md
pdaoust: Turns out that `.as_ref()` works: https://github.com/holochain/docs-pages/pull/547/commits/0d4b5de81259a2980cef89710b23b4d2849c1d0a#diff-1fb8e24f1d5eaf25c5f0ad6d38a2462809d6127c6e06ac8f8515e087c39a1f5dR221

src/pages/build/cryptography-functions.md
pdaoust: true. I pared down the example and the language that goes along with it: https://github.com/holochain/docs-pages/pull/547/commits/0d4b5de81259a2980cef89710b23b4d2849c1d0a

src/pages/build/cryptography-functions.md
pdaoust: did a lot of work on this in https://github.com/holochain/docs-pages/pull/547/commits/0d4b5de81259a2980cef89710b23b4d2849c1d0a

src/pages/build/cryptography-functions.md
pdaoust: I've pared down the examples significantly; I think they show the bare minimum required for making sense of box and secretbox without adding (too much) opinions about how they should be used in an app. Let me know what you all think: https://github.com/holochain/docs-pages/pull/547/commits/0d4b5de81259a2980cef89710b23b4d2849c1d0a

src/pages/build/cryptography-functions.md
pdaoust: rewrote this section, see https://github.com/holochain/docs-pages/pull/547#discussion_r2017139261

src/pages/build/cryptography-functions.md
pdaoust: This disappeared in this rewrite https://github.com/holochain/docs-pages/commit/0d4b5de81259a2980cef89710b23b4d2849c1d0a

src/pages/build/cryptography-functions.md
pdaoust: ha ha, good to know. Fixed in https://github.com/holochain/docs-pages/pull/547/commits/a73307bc13964368e454884922ebfcd2e3972571 where I also fix my mistake about the max hash length.

src/pages/build/cryptography-functions.md
pdaoust: I did a bit of rework... didn't remove it, but made it take up less space https://github.com/holochain/docs-pages/pull/547/commits/fa992a60df805edf9c6f23d4d2289b74edcac77c (It still makes me feel uncomfortable that these simple examples aren't wrapped in a `fn` tho)

src/pages/build/cryptography-functions.md
pdaoust: Did a little rewrite -- mentioned that this is a problem with almost all schemes these days. Thoughts? https://github.com/holochain/docs-pages/pull/547/commits/325c4ddc24cc5290896231aa95c8e1d5b201ab9b

src/pages/build/cryptography-functions.md
pdaoust: how's this? https://github.com/holochain/docs-pages/pull/547/commits/d1da5df43425792449b5c74ee743af17f35737de

src/pages/build/cryptography-functions.md
mattyg: I think its fine to wrap a simple example in a function, but I find this example overly complicated and noisy when we just want to show that calling the `hash_entry()` function gives you back its hash. Should be like 5 lines tops.

src/pages/build/cryptography-functions.md
mattyg: huh interesting

src/pages/build/cryptography-functions.md
mattyg: I'm good with the changes.

src/pages/build/cryptography-functions.md
pdaoust: How does it look now? https://github.com/holochain/docs-pages/pull/547/commits/33cf181619c5633e67c1d6dd0c8fc50a778d386a#diff-1fb8e24f1d5eaf25c5f0ad6d38a2462809d6127c6e06ac8f8515e087c39a1f5dL42-R45

src/pages/build/cryptography-functions.md
neonphog: I don't think you understood what I was suggesting. You still have a concept of a "GroupMessage" in here. You are still exposing the crypto functions as zome calls. You still are chaining calls together in a suggestion of how to do things.

```rust
fn example_shared_secret() {
    let _secret_ref = x_salsa20_poly1305_shared_secret_create_random(None).unwrap();
}

fn example_keypair() {
    let _pub_key = create_x25519_keypair().unwrap();
}

fn example_ingest_shared_secret() {
    let _secret_ref = x_salsa20_poly1305_shared_secret_ingest(
        [0; 32].into(),
        [1; 32].into(),
        XSalsa20Poly1305EncryptedData::new(
            [0; 24].into(),
            vec![],
        ),
        None
    ).unwrap();
}

// etc...
```

Fine if you guys want to overrule me and keep the patterns in here. But just so you understand what I'm voting for is *actually* deleting them.

src/pages/build/cryptography-functions.md
```diff
@@ -0,0 +1,363 @@
...
+// get all action/adventure movies.
+```
+
+!!! info Why would you hash an entry?
+Usually you have an entry hash and want to retrieve the entry data. When would you have data and want to hash it?
+
+Entries don't need to be written to the DHT in order to serve as a [base or target for links](/build/links-paths-and-anchors/#define-a-link-type). As long as the entry can be reconstructed by anyone who wants to store or retrieve links on its basis address, such as the tag in this example, its hash can also be reconstructed and used in `create_link` or `get_links`, even if there's no entry data at the address. This saves some storage and validation overhead for everyone.
+
+Or you may have just written an entry using `create` or `update` and want to use its hash in the same function call --- rather than retrieving the new written action by its hash to get the entry hash it contains, you can just hash the entry.
+!!!
```
c12i, line 54: I think this section here could be improved a little to answer the question `Why would you hash an entry?` a little more directly. Maybe bullet points might actually present the answers better?

src/pages/build/cryptography-functions.md
pdaoust: Okay, sounds like I did get it wrong -- I think I did understood correctly that you didn't want me giving opinionated advice on best practices that could get us into hot water, but I don't understand what your continuing concerns are? Specifically re:

* Exposing the host functions to the caller via a thin zome function wrapper, in what ways could this be a vulnerability? (I also think I've improved on that somewhat with the recent edit)
* Group messaging -- maybe I don't understand what your concern about that is, but the literature I've found on box and secretbox, including our own HDK Rustdoc, say that box is for authenticated two-party messages (using the crypto definition of message = payload, not message = chat) and secretbox is for multi-party messages.

src/pages/build/cryptography-functions.md
pdaoust: Is this closer to what you're thinking? https://github.com/holochain/docs-pages/pull/547/commits/c78883bdcf42eb82864cebee8afed88f4a53491b

c12i: Yea much better now

src/pages/build/cryptography-functions.md
pdaoust: I get you there for sure, and I really appreciate you looking out for the unknown unknowns that could wreck us. What I don't understand right now is that I believe that both the original and this rewrite are just cooking the noodles according the directions on the packet (with the exception of the opinionated-sounding naming on the original), and I've added lots of caveats about doing one's own research -- what remaining parts do you see as advice or my design?

neonphog: We are taking too much of both our time on this, so this will be my last comment on this PR. Please feel free to schedule a realtime meeting with me, or merge without following my advice.

As I've said, I think stringing more than one function call together is explicitly a design pattern. I don't know where you got the advice for group messaging from. If we're regurgitating advice from our own docs that doesn't satisfy my complaint. The source of truth is [the libsodium docs themselves](https://doc.libsodium.org/secret-key_cryptography/secretbox), and they tell you only how individual functions work, and do not recommend any patterns.

Once again for clarity: The functions themselves do not imply group messaging, only that the encryptor and decryptor must have the same secret key material.

pdaoust: Okay, I've got one more idea for decoupling this further, which should probably end up making things clearer to read too. Maybe I got the whole group thing from a bad source, I dunno. Anyhow, I'll rewrite and push, and the only feedback I'll ask for after this point is a yea/nay on the changes, so hopefully the PR blockers can be lifted and this can get merged.

neonphog: Sorry, I was responding to your pre-edited message...

> My concern about making things really abstract -- like just poking some literals into a function call -- is that it doesn't really communicate how to use it in any non-trivial codebase.

Yes, exactly. These are advanced enough topics, I don't think we should be telling people *how* to do things. I feel like they need to be doing their own real research into that to determine their own "how".

> but all of them need to know that you shouldn't send a symmetric key over an insecure channel, but hey, we also provide a way to create a secure channel for sending the key to each recipient.

There are two ways of going about this. You seem to want to string them all together in a very explicit pattern.

I feel it would be less explicit if we put a comment on the symmetric key saying, "This must be kept secret! Consider the design and security implications of using our `x_salsa20_poly1305_shared_secret_export` function."

Then separately have a usage example of that other function. That little bit of separation is what I'm looking for.

> But maybe the risk is so great that we shouldn't write about crypto functions at all, which is a point worth considering. Thoughts about that?

Hopefully my suggestion above shows a way I'd be comfortable with this being written. I'm also fine with just not writing about crypto functions.

pdaoust: okay, thank you for the follow-up. I independently started de-stringing the functions before I saw and read it, so hopefully it's in line with what you're calling for. I also turned them into private functions. Did the work in https://github.com/holochain/docs-pages/pull/547/commits/470ea7da91e5543addee95dc600f5e298e415d32 , https://github.com/holochain/docs-pages/pull/547/commits/46192bc1f6ba7a1bba27dff4859a28e8bc700864 , and https://github.com/holochain/docs-pages/pull/547/commits/4089af363dbbcc7220d0738f04d952b882159fc6

If this resolves your concerns, could you mark all the conversations as resolved? Don't know why, but the rules on this repo prevent PRs with open conversations from getting merged (I don't think this rule is true for other repos?)

neonphog: Thanks, looks good :+1:

matthme (approved):

mattyg (approved):

Issue #538, opened 2025-02-28
Practical docs for various host functions and stuff


pdaoust:
* `agent_info`
* `random_bytes`
* `sys_time`
* tracing
* `get_validation_receipts`

PR #548, opened 2025-03-27
Build guide: Cell Introspection

pdaoust:
This doesn't have a matching issue (although #538 does mention `agent_info`). Covers:

* `dna_info` (including the `#[dna_properties]` helper macro)
* `zome_info`
* `agent_info`
* `call_info`


src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+        // and its code intact; see
+        // https://developer.holochain.org/guide/cloning/
+        network_seed: _,
+        // Constants which can affect runtime behavior of this cell.
+        // They're usually specified as YAML and deserialized into a struct
+        // in your zome.
+        properties: _,
+        // The earliest valid timestamp for data in the network.
+        origin_time: _,
+        // A value used for tuning gossip, not useful for app development.
```
pdaoust, line 44: Should we create an issue to get rid of this field? Or will it just go away with kitsune2?

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    let AgentInfo {
+        // The public key, or agent ID, that the agent was initially using
+        // when they instantiated this cell.
+        agent_initial_pubkey,
+        // The agent's current ID.
+        agent_latest_pubkey,
+        chain_head,
+    } = agent_info()?;
+
+    // The agent's initial and latest public keys should be identical.
```
pdaoust, line 125: Is this useful for showing that the two pubkeys are the same if there's no DPKI? people will naturally ask which one they should use, and this kinda answers "either" -- or maybe that shouldn't be the answer

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+        // chain. Unlike AgentInfo.chain_head, this doesn't change as you
+        // try to write data in this function.
+        as_at,
+        // The capability grant that matches the claim used to call this
+        // function. See https://developer.holochain.org/build/capabilities/
+        // for more info.
+        cap_grant: _,
+    } = call_info()?;
+
+    // If this function is being called from another function in any cell in
```
pdaoust, line 175: Got some more assertions in this function; are they useful for showing what the data is about?

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    )?;
+    // Now, after writing to the chain, if we call agent_info again, the
+    // chain head in scratch space shouldn't match the persisted chain head.
+    assert_ne!(call_info()?.as_at, agent_info()?.chain_head, "we've written something now and the scratch space and persisted source chain are out of sync");
+
+    Ok(())
+}
+```
+
+!!! Be careful about tight coupling
```
pdaoust, line 203: Is this caveat useful? The encapsulation story with zomes and DNAs is such a hot mess, maybe it's better to stay silent about this.

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+<!--TODO: remove origin_time and quantum_time with 0.5 -->
+
+```rust
+use hdi::prelude::*;
+
+fn look_at_dna_info() -> ExternResult<()> {
+    let DnaInfoV2 {
+        // The human-readable name of the DNA, taken from the DNA manifest.
+        name: _,
+        // The hash of the DNA and its modifiers as defined *in this cell*.
```
mattyg, line 22:
```suggestion
        // The hash of the DNA's Integrity Zomes and its modifiers as defined *in this cell*.
```

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+To get information about the agent bound to the current cell, use the [`agent_info`](https://docs.rs/hdk/latest/hdk/info/fn.agent_info.html) host function. It takes no arguments and returns a result containing an [`AgentInfo`](https://docs.rs/hdk/latest/hdk/prelude/struct.AgentInfo.html) struct. _**Note**: This function is only available to coordinator zomes._
+
+<!-- TODO: talk about initial vs latest if DeepKey ever becomes real -->
+
+```rust
+use hdk::prelude::*;
+
+fn look_at_agent_info() -> ExternResult<()> {
+    let AgentInfo {
+        // The public key, or agent ID, that the agent was initially using
```
mattyg, line 117:
```suggestion
        // The public key of this agent
```

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+        // when they instantiated this cell.
+        agent_initial_pubkey,
+        // The agent's current ID.
```
mattyg, line 120:
```suggestion
        // The public key of this agent
        <!-- TODO remove this when it is deprecated -->
```

src/pages/build/cell-introspection.md
mattyg: Not useful and we're removing the field soon. I'd say delete this bit.

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    } = zome_info()?;
+    Ok(())
+}
+```
+
+## Get agent info
+
+To get information about the agent bound to the current cell, use the [`agent_info`](https://docs.rs/hdk/latest/hdk/info/fn.agent_info.html) host function. It takes no arguments and returns a result containing an [`AgentInfo`](https://docs.rs/hdk/latest/hdk/prelude/struct.AgentInfo.html) struct. _**Note**: This function is only available to coordinator zomes._
+
+<!-- TODO: talk about initial vs latest if DeepKey ever becomes real -->
```
mattyg, line 110: I believe the interface will change anyway.

src/pages/build/cell-introspection.md
mattyg: I don't think its useful.

src/pages/build/cell-introspection.md
mattyg: Yeah I think they're useful

src/pages/build/cell-introspection.md
pdaoust: I'll edit this in a separate commit because there are some other things that need to change if this is true.

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,216 @@
...
+title: "Cell Introspection"
+---
+
+::: intro
+Holochain lets you get details about parts of the cell and its current state --- the current zome and DNA, the context of the current function call, and the agent bound to the hApp's cells.
+:::
+
+## Get DNA info
+
+To get details about the DNA that a function is executing in, use the [`dna_info`](https://docs.rs/hdk/latest/hdk/prelude/fn.dna_info.html) host function, which is available to both coordinator and integrity zomes. It takes no inputs and gives back a result containing a [`DnaInfoV2`](https://docs.rs/hdk/latest/hdk/prelude/struct.DnaInfoV2.html) struct. See the [DNAs page](/build/dnas/) for more info on the values in tis struct.
```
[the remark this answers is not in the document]
pdaoust: GH wouldn't let me accept this edit. Did it in https://github.com/holochain/docs-pages/pull/548/commits/324dde1cff3c4862ae8cb4e88fed4fd7701e3584

src/pages/build/cell-introspection.md
pdaoust: changed in https://github.com/holochain/docs-pages/pull/548/commits/3be1ac63a622b1aa6f6a5e6d361711d5a90e1d6e

src/pages/build/cell-introspection.md
pdaoust: changed in https://github.com/holochain/docs-pages/pull/548/commits/3be1ac63a622b1aa6f6a5e6d361711d5a90e1d6e

src/pages/build/cell-introspection.md
pdaoust: changed in https://github.com/holochain/docs-pages/pull/548/commits/3be1ac63a622b1aa6f6a5e6d361711d5a90e1d6e

src/pages/build/cell-introspection.md
pdaoust: removed in https://github.com/holochain/docs-pages/pull/548/commits/af2ad43e54a9d54df817a927c247d458338b1f17

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,209 @@
...
+        _latest_action_sequence_index,
+        _latest_action_timestamp,
+    ) = chain_head;
+
+    Ok(())
+}
+```
+
+!!! info New actions aren't persisted until after a function call completes
+At the start of every function call that can write data to the source chain, Holochain [creates a new 'scratch space'](/build/zome-functions/#atomic-transactional-commits) to store new writes, then flushes those writes to the source chain _after the function call completes and all writes pass validation_. That means that the `chain_head` property reflects the current state of the source chain _in the scratch space_, not the state of the source chain persisted to storage. Validation failure or hardware failure can cause the scratch space to be rolled back or lost.
```
matthme, line 141: > That means that the `chain_head` property reflects the current state of the source chain _in the scratch space_, not the state of the source chain persisted to storage

I didn't know that (and also never used the chain_head field so I never had to think about it). I wonder whether we should also add this to the Rust docs of the AgentInfo struct? Could potentially be sneaked in together with the PR to remove the `agent_latest_pubkey` field...

src/pages/build/cell-introspection.md
```diff
@@ -0,0 +1,209 @@
...
+!!!
+
+## Get info about the call context
+
+To get information about the context of the currently executing call, use the [`call_info`](https://docs.rs/hdk/latest/hdk/info/fn.call_info.html) host function. It takes no arguments and returns a result containing a [`CallInfo`](https://docs.rs/hdk/latest/hdk/prelude/struct.CallInfo.html) struct. _**Note**: This function is only available to coordinator zomes._
```
matthme, line 146: > _**Note**: This function is only available to coordinator zomes._

Interesting...didn't know that either. I thought that integrity zomes can in principle also contain zome functions so I wonder why that's the case.

matthme (approved):

src/pages/build/cell-introspection.md
pdaoust: It turns out I made that up (I thought I'd read it somewhere), but it also turns out I'm accidentally right! On the host side, `agent_info` calls [these functions](https://github.com/holochain/holochain/blob/develop/crates/holochain_state/src/source_chain.rs#L656-L666), which appear to add the scratch space onto the as_at state.

pdaoust: okay, I put up a PR to document this. https://github.com/holochain/holochain/pull/4841

mattyg (dismissed):

mattyg (approved):

PR #554, opened 2025-04-02
fix: callout rendering

mattyg:


pdaoust (approved):
Thanks for catching this!

PR #568, opened 2025-05-22
Edit: clearer advice in migration guide for flakes that don't look the same

pdaoust:
* Update the advice to say that, if your flake file looks different, just copy the example
* Change behaviour of the code copy button so that it only copies unchanged or added lines in a diff


src/client-side/index.ts
```diff
@@ -1,7 +1,26 @@
 function copyCodeBlockClickHandler(e: Event) {
   const button = e.target as HTMLButtonElement;
   const codeEl = button.parentElement?.querySelector("code");
-  navigator.clipboard.writeText(codeEl?.innerText || "")
+  if (!codeEl) { return; }
```
matthme, line 4: sweet.

matthme (approved):
Looks great :+1:

src/client-side/index.ts
pdaoust: Thanks! :)

Issue #444, opened 2024-03-29
Better syntax highlighting for rust code


mattyg:
The syntax highlighting we're using in code examples doesn't clearly distinguish many of the syntax elements of rust code. This makes it far less readable.

For example, this is our docs:
![image](https://github.com/holochain/docs-pages/assets/159270/cc067c60-a80e-4d9d-9358-20b51767b14e)

And this is vscode:
![image](https://github.com/holochain/docs-pages/assets/159270/3e066f4e-ce82-4c4b-b1b2-f4693395ee3a)

mattyg:
I'm guessing it's the theme used by the syntax highlighting library that just needs to have different colors for more code elements. Some themes may not colorize all elements.

pdaoust:
could be -- although it may be that highlight.js doesn't have an expressive enough grammar for this. Here are two promising candidates:

* https://github.com/wooorm/starry-night
* https://github.com/shikijs/shiki

Criteria:

* JS
* Easy to integrate with 11ty (ideal situation is a sync function that takes a code block string and a language and returns HTML)

pdaoust:
Testing: Here's Shiki, which claims it keeps parity with VSCode but can be used server-side. Using Dark Plus theme. A dead ringer for your VSCode screenshot.

![Image](https://github.com/user-attachments/assets/b4ce8c58-fed4-49c2-aabd-e5ff12cf5f3e)

pdaoust:
Might be addressed by #574 . It looks like Docusaurus hard-codes Prism.js and there's little to no documentation on how to extend Docusaurus to, say, integrate other code highlighters. But either we were using the wrong Prism theme, or it didn't do really well one of the languages we use a lot of (probably Svelte). Anyhow, it'll be interesting to see what that spike uncovers.

PR #569, opened 2025-05-23
Feat/fix: Get diff highlighing working and ergonomic

pdaoust:
* Fix broken diff+language highlighting so diffs show highlighted code
* UI to toggle between diff and applied diff and copy whichever you like
* Edits to content to remove unnecessary "your edited thing should look like this", since it's in the 'Final' tab of a diff code block now.


matthme (approved):

pdaoust:
@jost-s thank you! I don't think there is an off-the-shelf package -- I did a quick search and didn't really find anything, but given that diff+syntax highlighting and the copy button are both add-on features (one of which was written by Matthew N), and highlight.js isn't as popular as Prism, I think the combinatorial likelihood of it existing was pretty slim to begin with :)

We're considering moving to a different code highlighter like Prism (and a different static site generator) in the future due to #444 and #574 , so I recognise that all this custom code may be a liability there... or I may be able to reuse it!

PR #577, opened 2025-06-05
Chore: new infra URLs in 0.5 upgrade guide

### LINKED ISSUE #576: Document new dev-test servers on 0.5 upgrade guide
With kitsune2 comes new networking infra servers. Document them in the upgrade guide.

pdaoust:
Closes #576 . Shows how to modify a kangaroo config file and recommends that devs maintain their own infrastructure.


src/pages/resources/upgrade/upgrade-holochain-0.5.md
```diff
@@ -362,6 +363,59 @@ The `NetworkInfo` endpoint of the app API has been removed, which means the `App
 
 **Note: This is an advanced feature that you'll only encounter if you're building a custom runtime that tries to get diagnostic data.** The return value of the [`AgentInfo`](https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AdminRequest.html#variant.AgentInfo) endpoint in the admin API has changed. Currently it outputs a vector of JSON-serialized [`kitsune2_api::AgentInfoSigned`](https://docs.rs/kitsune2_api/latest/kitsune2_api/struct.AgentInfoSigned.html) values.
 
+## Production apps: new network infrastructure servers
+
+With the change to Kitsune2 Holo is retiring their public bootstrap and signal servers. For testing, we're offering public servers you can use. **We request and recommend that you maintain your own bootstrap and signal servers**<!-- TODO: document this https://github.com/holochain/docs-pages/issues/573 -->, as these servers are rate-limited and have no uptime guarantees.
+
+If you're bundling your hApp with [Kangaroo](https://github.com/holochain/kangaroo-electron), you'll need to add to your `kangaroo.config.ts` file:
```
matthme, line 370: Well, if they're already using kangaroo for earlier holochain versions, then they'll need to change more than just the config here. They'd have to merge in upstream changes into their fork of kangaroo that also includes the new conductor config template for example and associated logic. So maybe we should rephrase this a bit here and say that, after updating their fork to use the 0.5 branch of kangaroo that they have to replace the default urls in that config with their own servers.

src/pages/resources/upgrade/upgrade-holochain-0.5.md
```diff
@@ -362,6 +363,65 @@ The `NetworkInfo` endpoint of the app API has been removed, which means the `App
 
 **Note: This is an advanced feature that you'll only encounter if you're building a custom runtime that tries to get diagnostic data.** The return value of the [`AgentInfo`](https://docs.rs/holochain_conductor_api/latest/holochain_conductor_api/enum.AdminRequest.html#variant.AgentInfo) endpoint in the admin API has changed. Currently it outputs a vector of JSON-serialized [`kitsune2_api::AgentInfoSigned`](https://docs.rs/kitsune2_api/latest/kitsune2_api/struct.AgentInfoSigned.html) values.
 
+## Production apps: new network infrastructure servers
+
+With the change to Kitsune2 Holo is retiring their public bootstrap and signal servers. For testing, we're offering public servers you can use. **We request and recommend that you maintain your own bootstrap and signal servers**<!-- TODO: document this https://github.com/holochain/docs-pages/issues/573 -->, as the test servers are rate-limited and have no uptime guarantees. You can find the server binary on [crates.io](https://crates.io/crates/kitsune2_bootstrap_srv) or adapt [this example `Dockerfile`](https://github.com/holochain/kitsune2/blob/main/docker/kitsune2_bootstrap_srv/Dockerfile).
```
ThetaSinner, line 368: > With the change to Kitsune2 Holo is retiring their public bootstrap and signal servers

I find this a little misleading. The old servers are still running and Holo is still using them. They aren't imminently going away. But the new bootstrap/sbd is different software and won't work with the old servers. I think sbd might work but I'm not sure about that. For sure bootstrap is a different API.


I'd prefer to present it that way, that it's a new service and Holo isn't providing anything yet. They might at some point, but it probably won't be public, it'll be provided as part of the hApp hosting service.

ThetaSinner, line 368: Should the dev test servers be mentioned more explicitly here? Rather than just being removed lines below? They won't have existed in the old kangaroo configuration for people to remove.

pdaoust: oh! I was not aware of all that; I'll update.

pdaoust: Right, and initially the diff had shown the dev test server URLs being _added in_ to a config file, but then @matthme pointed out that (a) that doesn't work because the whole config file will need changes, and (b) maybe it's better to show the replacement of those lines (which will exist after following the advice in https://github.com/holochain/docs-pages/pull/577/files#diff-a929c5cfeaf326697a5b7d2204d4df391d0ca6a285a2760592013e1c903b9191R372-R374 with one's own production servers).

Thoughts on that?

matthme (approved):

PR #582, opened 2025-06-23
fix: info block markdown formatting

matthme:
Fixes a small markdown formatting issue.


pdaoust (approved):
looks good; thanks for catching it!

PR #584, opened 2025-06-30
Chore: allow 'edit' commit type

pdaoust:
Just getting annoyed with cocogitto disliking my most common commit type, so I'm allowing it for this repo.


pdaoust:
turns out cocogitto-bot doesn't honour that file; closing.

PR #625, opened 2025-12-15
fix: warrants are stable now, edit documentation

### LINKED ISSUE #622: [BUG] Docs mention warrants as future feature
## What is incorrect

This somehow got missed in the 0.6 docs sweep -- https://developer.holochain.org/build/must-get-host-functions/ has a callout at the bottom of the page that says:

> In the future, we intend to introduce ‘warrants’, a feature which will allow validators to communicate failures to each other for related data. Until then, if any of your validation code uses must_get_valid_record to retrieve a dependency, we recommend that the dependency’s validation code for the StoreRecord operation cover all possible checks.

Now that warrants are stable, we need to remove the 'in the future' bit, but we also need to warn devs that, while warrants may be generated for any sort of op, they're only communicated to the record's AAA, not among all validators of the record's ops.

pdaoust:
Closes #622 . Now that warrants are stable, the documentation needs to reflect that. The info box also needed to change because warrants really don't have much to do with `must_get_*` functions, and the explanation of the limits of `must_get_valid_record` was itself too limited.

I also took the opportunity to edit a code sample on that page, which can be much simpler and more network-efficient now that you can `must_get_agent_activity` by time bounds.

## Summary by CodeRabbit

* **Documentation**
  * Improved DHT operation anchors and cross-references for more reliable navigation.
  * Clarified host-function behavior around timestamp windows and bounded activity queries.
  * Expanded guidance on validation expectations and where to enforce strong guarantees.
  * Enhanced explanation of warrants and added clearer links in querying guides for easier lookup.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


src/pages/build/must-get-host-functions.md
```diff
@@ -169,9 +165,11 @@ fn check_that_action_exists_and_is_valid_and_has_valid_public_app_entry(action_h
 ```
 
 !!! info This may not catch all validation failures
-`must_get_valid_record` checks for validation success or failure on the [`StoreRecord` DHT operation](/build/dht-operations/#store-record). Validation code for other DHT operations produced from the same action may have executed and failed.
+`must_get_valid_record` checks for validation success or failure on the [`StoreRecord` DHT operation](/build/dht-operations/#store-record). Validation code for other DHT operations produced from the same action (such as [`RegisterUpdate`](/build/dht-operations/#register-update) or [`RegisterDeleteLink`](/build/dht-operations/#register-delete-link)) may have failed, but will not reflect on the record's validity.
```
ThetaSinner: True, though that's not how the system was supposed to work. This sits inside the data model re-work discussion. Fine for now, but we need to remember to come back and update this with 0.7

PR #641, opened 2026-05-16
fix: remove local artifact that was accidentally checked in

pdaoust:


pdaoust:
Merged even with build errors because they were spurious -- 403 forbidden errors on www.holochain.org URLs 🤷🏼‍♂️

PR #644, opened 2026-05-28
chore: remove hello-world step from getting started tutorial

pdaoust:
### NOTE: this PR is ready to go, but we won't deploy it until Holochain 0.7 is the recommended version, because the changes to the scaffolding tool won't apply until then.

Things I did:

* remove the hello world page (step 2) from the tutorial
* removed its image assets
* renamed steps 3 (forum) and 4 (packaging) to 2 and 3 respectively
* renamed image assets for step 3
* updated menu entries
* copied still-useful content from the hello world page onto the forum page
* added 301 redirects to the new forum and packaging page URLs

Closes #631


src/pages/get-started/2-forum-app-tutorial.md
```diff
@@ -779,13 +825,32 @@ hc scaffold collection --help
...
+
+Then load the new profile by running the command:
 
 ```shell
-sudo chown root:root node_modules/electron/dist/chrome-sandbox && sudo chmod 4755 node_modules/electron/dist/chrome-sandbox
+sudo apparmor_parser -r /etc/apparmor.d/hc-spin-electron
 ```
 
-You'll need to do this once (but only once) for every new project you scaffold. You can find out more [here](/get-started/install-advanced/#fixing-the-suid-sandbox-error-in-ubuntu-24-04-and-later).
+You'll need to do this once for your system, then other projects you scaffold should just work. You can find out more [here](/get-started/install-advanced/#fixing-the-suid-sandbox-error-in-ubuntu-24-04-and-later).
```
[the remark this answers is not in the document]
pdaoust: the suggestion is just plain wrong and doesn't address the issue raised.

src/pages/get-started/2-forum-app-tutorial.md
jost-s: These are the only ones left
```suggestion
❯ svelte
 headless (no ui)
```

jost-s (dismissed):
Looks great! Just one change requested regarding the UI framework selection.

src/pages/get-started/2-forum-app-tutorial.md
pdaoust: @jost-s ah, I was wondering if you'd already removed the other UI frameworks. How about the package managers -- is it reduced down to npm now too?

[EDIT: I tried to check for myself, but it looks like the one you get through holonix 0.6 is still showing all the frameworks so I assume any changes to the package managers won't be showing up yet either.]

jost-s: @pdaoust Wait, now I see. This is the `main` branch, so it's for Holochain v0.7. We've left everything in place for 0.6. This removal only affects 0.7. There should be just `npm` as pm, `svelte` as UI and `headless` as non-UI template.

pdaoust: @jost-s okay -- I knew it wasn't ready for release yet (didn't know which release it was synced with) and was assuming we'd just deploy it when the release in question is ready :)

pdaoust: ack, this is what comes of not checking -- we dropped the alternate package managers in `main-0.6` already. Never mind 😅

ThetaSinner (approved):

jost-s:
> NOTE: this PR is ready to go, but we won't deploy it until Holochain 0.7 is the recommended version, because the changes to the scaffolding tool won't apply until then.

src/pages/build/connecting-a-front-end.md
```diff
@@ -12,7 +12,7 @@ The most important thing to remember about a hApp is that it runs in the Holocha
 
 This interface is **only exposed to processes on the local device**, not to external network adapters. This helps prevent unauthorized agents from accessing the hApp. It also means the front end must be distributed with the hApp and a Holochain runtime.
 
-Some Holochain runtimes bundle the conductor and a front end host that serves your HTML/JavaScript-based runtimes. Take a look at the [Packaging and Distribution](/get-started/4-packaging-and-distribution/) page from the Get Started guide for the choices.
+Some Holochain runtimes bundle the conductor and a front end host that serves your HTML/JavaScript-based runtimes. Take a look at the [Packaging and Distribution](/get-started/3-packaging-and-distribution/) page from the Get Started guide for the choices.
```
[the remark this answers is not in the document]
ThetaSinner: Not actioning this one.

The only edit this PR makes to line 15 is the renumbering `/get-started/4-packaging-and-distribution/` -> `/get-started/3-packaging-and-distribution/`. "front end host" is pre-existing text that the PR doesn't touch.

There's also no repo convention to enforce here: `src/pages` contains 38 unhyphenated "front end" and 26 hyphenated "front-end", frequently on the same page. Hyphenating this single instance would make the repo marginally less consistent, not more. If we want to standardise on one form it belongs in its own sweep.

src/pages/get-started/2-forum-app-tutorial.md
```diff
@@ -779,13 +821,32 @@ hc scaffold collection --help
...
+}
+EOF
+```
+
+Then load the new profile by running the command:
 
 ```shell
-sudo chown root:root node_modules/electron/dist/chrome-sandbox && sudo chmod 4755 node_modules/electron/dist/chrome-sandbox
+sudo apparmor_parser -r /etc/apparmor.d/hc-spin-electron
 ```
```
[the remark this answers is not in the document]
ThetaSinner: The observation is fair, but it isn't in scope for this PR and I'd rather not change it here.

That profile line is not new. It is byte-identical to the one already published in `2-hello-world.md`, which this PR deletes:

```
profile hc-spin-electron /**/node_modules/electron/dist/electron flags=(unconfined) {
```

It shows up as an addition only because the Ubuntu 24.04 workaround moved into the forum tutorial when the hello-world page went away. The guidance is live on the site today, so this PR changes nothing about the exposure.

On the substance: the glob is deliberate. The docs can't know where a reader will scaffold their project, so the attachment path has to match `node_modules/electron/dist/electron` wherever it lands. Narrowing it means either knowing the project path in advance or dropping `flags=(unconfined)`, and either would need testing that the workaround still does its job on 24.04.

That's a call for whoever owns the Ubuntu workaround, not something to change silently in a PR whose purpose is renumbering the getting-started steps. Worth a separate issue.

src/pages/get-started/index.md
```diff
@@ -119,6 +119,6 @@ Congratulations! The Holochain development environment is now set up successfull
 
 ### Next up
 
-Now it's time to download, compile, and try an example Holochain application.
+Now it's time to create your own Holochain application.
 
-[Try 'Hello World' →](/get-started/2-hello-world/){.btn-purple}
\ No newline at end of file
+[Build a forum hApp →](/get-started/2-forum-app-tutorial/){.btn-purple}
```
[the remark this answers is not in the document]
ThetaSinner: Correct that the file has no trailing newline, and MD047 is enabled (`.markdownlint.json` only overrides MD007). But it's a pre-existing repo-wide condition rather than anything this PR introduces: **46 of 73** markdown files under `src/` end without one, and the diff shows `\ No newline at end of file` on both the old and new side of that last line.

markdownlint also isn't wired into CI, so nothing enforces the rule today.

Fixing this one file would leave the other 45 untouched and imply an enforcement that doesn't exist. If the rule is wanted, the useful change is adding markdownlint to CI and sweeping all 46 in one go - happy to raise that separately.

PR #31, opened 2019-08-01
add note about dropping the `v` from release tags

pdaoust:

Issue #17, opened 2019-02-21
developer hdk docs, the upper left link to the top level of the docs is broken


dhtnetwork:
In the developer hdk docs, the upper left link to the top level of the docs is broken: It should be something like https://developer.holochain.org/api/latest/hdk but instead points to https://developer.holochain.org/api/latest/index.html.

dhtnetwork:
HDK docs point to the latest version.

PR #647, opened 2026-07-27
docs: add Holochain 0.6 → 0.7 upgrade guide

ThetaSinner:
## Summary

- Adds the **0.6 → 0.7 upgrade guide** at `/resources/upgrade/upgrade-holochain-0.7/`, following the structure of the 0.6 guide.
- Adds a **0.7 compatibility table**, marked as a release candidate.
- Wires both pages into the nav and resource indexes, and repoints the `/latest/` redirects.

Written against **Holochain 0.7.0-rc.4** (hdi 0.8.0-rc.3 / hdk 0.7.0-rc.3 / `@holochain/client` 0.21.0-rc.1 / hc-spin 0.700.0-rc.1). Every version number that will need bumping at GA carries a `TODO(upgrade)` comment.

### Sources

Content is drawn from the reference upgrades — dino-adventure [#17](https://github.com/holochain/dino-adventure/pull/17), [#24](https://github.com/holochain/dino-adventure/pull/24), [#25](https://github.com/holochain/dino-adventure/pull/25), [#29](https://github.com/holochain/dino-adventure/pull/29), [#31](https://github.com/holochain/dino-adventure/pull/31) and scaffolding [#576](https://github.com/holochain/scaffolding/pull/576) — plus an audit of all 61 `BREAKING` entries in the 0.7 range of Holochain's changelog. Signatures were verified against the source at the `holochain-0.7.0-rc.4` tag rather than reconstructed from diffs.

Note that #29/#31 alone don't cover 0.6 → 0.7: the app went to 0.7.0-dev.6 and dev.23 first, so the `wasmer_sys` → `wasmer-sys-cranelift` rename and the `TransportStats` → `ApiTransportStats` change come from the earlier PRs.

### What the guide covers

- **The action model rewrite** — `Action` as `header` + `data`, `TypedAction<D>`, renamed `FlatOp` variants, and the removal of the per-variant action structs and `EntryCreationAction`. Two mapping tables plus diffs for the validation and `signal_action` functions people hand-edit.
- **tx5/WebRTC removal** — iroh is the only transport; `signal_url` and `webrtc_config` are gone and now hard-error.
- **No migration path** — DNA hashes change and databases were renamed, so existing installs must be cleared.
- Other HDK/HDI changes, Rust and JavaScript client changes, and the conductor config diff for Kangaroo-based projects.

### Points for review

- **The `/latest/` redirects now point at the 0.7 pages.** The comment in `netlify.toml` says to do this only once a release is *recommended*, so this may be premature while 0.7 is an RC — happy to drop that commit.
- **0.6 keeps its "recommended" badge** and the 0.7 table is labelled "Release candidate", since flipping 0.6 to maintenance now would leave no recommended version. Both carry `TODO(upgrade)` notes for the switch at GA.
- **No `main-0.7` Holonix branch exists yet**, so the `flake.nix` diff shows `main-0.7` with an admonition pointing readers at `main` in the meantime.
- **No 0.7 scaffolding release yet** (holochain/scaffolding#576 is still open), so the "update the `validate` body" section points at dino-adventure's integrity zome instead of recommending a re-scaffold.

### Also in this PR: a fix for the site-wide link check

The Netlify deploy has been failing on **every** branch since 20 May, including production deploys from `main` — the `checklinks` plugin was hitting three dead links unrelated to this guide:

| Link | Cause |
|---|---|
| `tryorama/blob/main-0.5/…appwithoptions.md` | `main-0.5` branch deleted |
| `tryorama/blob/v0.18.1/…installapp.md` | `v0.18.1` tag deleted; only `v0.19.2` remains |
| `network-services/blob/main/dev-test/docker-compose.yaml` | directory renamed to `dev-test-bootstrap2`, compose file replaced by a cloud-init unit |

The last commit repairs all three, in `build/testing-with-tryorama.md` and `resources/upgrade/upgrade-holochain-0.5.md`. Both Tryorama links now point at `main`, matching the neighbouring links in the same files; the network-services link points at the equivalent lines of `dev-test-bootstrap2/cloud-init.yaml`, which still show the TLS cert and key configuration the surrounding text describes. This unblocks production deploys as a side effect. Happy to split it out if you'd rather keep this PR to the guide.

Builds clean, spell-checks clean, and a full local run of the `checklinks` plugin over the whole site reports no failures.


src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+To upgrade your hApp written for Holochain 0.6, follow these steps:
+
+1. Update your `flake.nix` to use the 0.7 version of Holochain: {#update-nix-flake}
+
+    ```diff
+     {
+       description = "Flake for Holochain app development";
+       inputs = {
+    -    holonix.url = "github:holochain/holonix?ref=main-0.6";
+    +    holonix.url = "github:holochain/holonix?ref=main-0.7";
```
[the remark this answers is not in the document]
ThetaSinner: Addressed in 84145cb.

Confirmed `main-0.7` does not exist — `holochain/holonix` currently has `main`, `main-0.2` through `main-0.6`, and `update-to-0.7.0-rc.0`. The diff now uses `ref=main`, and the admonition below it is inverted: it explains that `main` tracks the release candidates and to switch to `main-0.7` once 0.7.0 ships. The `TODO(upgrade)` comment above it now covers changing the ref as well as removing the note.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+    ```diff:json
+     {
+         "devDependencies": {
+    -        "@holochain/hc-spin": "^0.601.3",
+    +        "@holochain/hc-spin": "^0.700.0-rc.1",
+             "concurrently": "^6.5.1",
+             "get-port-cli": "^3.0.0"
+         }
+     }
+    ```
```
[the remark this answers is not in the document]
ThetaSinner: Not actioned — this matches the existing convention in the sibling guide.

`markdownlint-cli2` reports exactly 3 MD031 violations in `upgrade-holochain-0.6.md` and exactly 3 in `upgrade-holochain-0.7.md`, in the same place: a closing fence indented inside an ordered list item, immediately followed by the next item marker. This PR does not introduce a new pattern.

markdownlint is also not run in CI — the repo has a `.markdownlint.json`, but it only sets `MD007.indent` for editors, and the only workflow is `spellcheck.yaml`.

Rendering is unaffected either way: the list already contains blank lines within item content, so it is loose, and 11ty renders it as `<ol><li><p>…`. Adding blank lines after the closing fences would change nothing visually while diverging from the 0.6 guide.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+
+Sometimes dependency updates break builds! Cargo updates dependencies to the most recent versions that are _claimed_ to be compatible. However, sometimes package maintainers accidentally introduce a breaking change without labelling it as such. If you encounter build errors, this may be the cause. If this happens, try locking the incompatible dependency to a prior version in your `Cargo.toml` file, running `cargo update` again, and see if that fixes it.
+
+## Rust client and test changes
+
+These only apply if you use the Rust `holochain_client` crate, or write Sweettest tests that sign data directly.
+
+### `dump_network_stats` returns `HolochainTransportStats`
+
+The admin and app websockets used to return different types --- `kitsune2_api::ApiTransportStats` and `kitsune2_api::TransportStats` respectively, with the app response missing blocked message counts. Both now return `HolochainTransportStats`, which also converts Kitsune2 `Space` values into `DnaHash` values so you can use them directly.
```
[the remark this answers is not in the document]
ThetaSinner: Already fixed in b071cbd, before this round. Both occurrences now read "WebSocket clients", matching the spelling used elsewhere in the docs (`resources/index.md`, `build/connecting-the-parts.md`, `resources/glossary.md`) rather than adding `websockets` to the dictionary. The `spellcheck` job is green.

ThetaSinner:
Replying here to the outside-diff comment on `netlify.toml` (195-206), since it was not posted inline.

**Partially actioned in 1397615.** The inconsistency is real, and I had already flagged it in the PR description. I have kept the redirects on 0.7 and made the comment match, rather than reverting them to 0.6.

Reason: pointing `/latest/` at the RC is deliberate here — while 0.7 is the release people are being pointed at to try, `/resources/upgrade/latest/` is the URL that gets handed out, and sending them to the 0.6 guide defeats the purpose of the alias. The comment now says that explicitly:

```
# While a release is still a release candidate, these may point at its pages ahead of
# `recommended` status, because the RC is what people asking for the latest are being
# pointed at.
```

The existing `TODO(upgrade)` above it still covers the bump at GA.

This is a judgement call rather than a correctness issue, and it is isolated in its own commit, so reverting to 0.6 is one commit away if the maintainers prefer the stricter reading of the contract.

src/pages/resources/compatibility/holochain-0.7.md
```diff
@@ -0,0 +1,32 @@
...
+| [`hdk` library](https://crates.io/crates/hdk)                                                     | [0.7.0-rc.3](https://crates.io/crates/hdk/0.7.0-rc.3)                                           |
+| [`hdi` library](https://crates.io/crates/hdi)                                                     | [0.8.0-rc.3](https://crates.io/crates/hdi/0.8.0-rc.3)                                           |
+| [JavaScript client library](https://github.com/holochain/holochain-client-js)                     | [0.21.0-rc.1](https://www.npmjs.com/package/@holochain/client/v/0.21.0-rc.1)                    |
+| [Rust client library](https://github.com/holochain/holochain/tree/main/crates/client)             | [0.9.0-rc.4](https://crates.io/crates/holochain_client/0.9.0-rc.4)                              |
+| [`kangaroo-electron` app template repo](https://github.com/holochain/kangaroo-electron)           | Not yet updated for 0.7                                                                         |
```
ThetaSinner: It has been updated for 0.7.0-rc.3, it'd just be `main` for now. Are you waiting on a `main-0.7` branch?

src/pages/resources/upgrade/upgrade-holochain-0.5.md
```diff
@@ -317,7 +317,7 @@ If you want to use these features, [build a custom Holochain binary](https://git
...
+**Note: This information is only relevant if you're building a runtime or using Tryorama's [`Conductor.prototype.installApp`](https://github.com/holochain-open-dev/tryorama/blob/main/docs/tryorama.conductor.installapp.md) method, either directly or through a helper like `scenario.addPlayersWithApps`.** The `Bundle` option (deprecated in 0.4.2) is now removed from [`AppBundleSource`](https://docs.rs/holochain_types/0.5.2/holochain_types/app/enum.AppBundleSource.html). If you need to pass bundle bytes to the admin API's `InstallApp` endpoint, use [`AppBundleSource::Bytes`](https://docs.rs/holochain_types/0.5.2/holochain_types/app/enum.AppBundleSource.html#variant.Bytes) (introduced in 0.4.2) and pass the bytes of an entire hApp bundle file instead.
```
ThetaSinner, line 320: Probably a bad idea to point at latest main for `0.5` Holochain - were the tags not carried over to holochain-open-dev when it was forked?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
+---
+title: Holochain Upgrade 0.6 → 0.7
+---
+
+::: intro
+For existing hApps that are currently using Holochain 0.6, here's the guide to get you upgraded to 0.7.
+
+There are three big changes in Holochain 0.7:
+
+* **The action model has been rewritten.** An `Action` is now a `header` (the fields every action shares) plus a `data` payload. The per-variant action structs and `EntryCreationAction` are gone. This touches every integrity zome's `validate` callback and every coordinator zome's `signal_action` function, and it's the bulk of the upgrade work.
```
ThetaSinner: I thought `EntryCreationAction` ended up getting replaced with something else? Maybe that hasn't been released yet.

See `EntryCreationData` which should be available on `holochain/holochain` on the `develop` branch. If that's not released yet that type, then let's leave a comment behind for ourselves here.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
++holochain = { workspace = true, features = ["wasmer-sys-cranelift", "test_utils"] }
+ tokio = { workspace = true }
+```
+
+A number of crates have also removed the implicit Cargo features that came from enabling an optional dependency. If any of your crates enable features on Holochain crates directly, these are the renames most likely to affect you:
+
+| Crate | Removed feature | Use instead |
+|---|---|---|
+| `holo_hash` | `serde`, `serde_bytes` | `serialization` |
+| `holo_hash` | `rusqlite` | `sqlite` or `sqlite-encrypted` |
```
ThetaSinner: Didn't those feature flags later get removed because `holo_hash` no longer depends on `rusqlite`?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+Import from a prelude instead:
+
+```diff:rust
+-use holochain_zome_types::action::Action;
+-use holochain_integrity_types::Entry;
++use holochain_zome_types::prelude::Action;
++use holochain_integrity_types::prelude::Entry;
+```
+
+### The action model has changed
```
ThetaSinner, line 188: Might be helpful to say something in this section about how the data structs largely carry the same fields as the old action structs used to, once the header fields are pulled out. A few fields are renamed for consistency but that's how I think about this refactor. I think it helps build intuition about where to look for fields that have moved?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+}
+```
+
+`ActionHeader` holds `author`, `timestamp`, `action_seq` and `prev_action`. Everything else lives in the `ActionData` variant.
+
+Use this table to rename types in your zome code:
+
+| Holochain 0.6 | Holochain 0.7 |
+|---|---|
+| `Action::Create(c)`, `Action::Update(u)`, … | `ActionData::Create(c)`, `ActionData::Update(u)`, … — matched on `action.data` |
```
ThetaSinner: What are all these "..." about in this table?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+Because that data is now directly available, the fields that used to be copied out alongside the action are gone. Read them from `action.data` instead:
+
+| Removed field | Read instead |
+|---|---|
+| `original_action_hash` on `OpRecord::UpdateEntry` | `action.data.original_action_address` |
+| `original_action_hash` on `OpRecord::DeleteEntry` | `action.data.deletes_address` |
+| `original_action_hash` on `OpRecord::DeleteLink` | `action.data.link_add_address` |
+| `base_address`, `target_address`, `tag` on the link ops | `action.data.base_address`, `.target_address`, `.tag` |
+
+`OpUpdate::original_action_hash()` and `original_entry_hash()` remain as accessor methods, as do `agent()`, `new_key()` and `original_key()` on `OpEntry`, `OpRecord` and `OpActivity`.
```
ThetaSinner: This information changes again after PR https://github.com/holochain/holochain/commit/3c39e86eccd90bb7f142749551bcd00ace6ccff1

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+ ) -> ExternResult<ValidateCallbackResult> {
+     Ok(ValidateCallbackResult::Valid)
+ }
+```
+
+Anywhere your validation logic reads a common action field, go through `header`:
+
+```diff:rust
+-if &action.author != record.action().author() {
++if &action.header.author != record.action().author() {
```
ThetaSinner: Isn't there an `action.author()` here for convenience?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+     return Err(wasm_error!(WasmErrorInner::Guest(
+         "Only the author can link their own post".to_string()
+     )));
+ }
+```
+
+### Update the `validate` callback body
+
+The body of `validate` is scaffolded code that dispatches to the functions above, so most of this work is mechanical. Rather than rewriting it by hand, work from [`dino_adventure`'s integrity zome](https://github.com/holochain/dino-adventure/blob/main/dnas/dino_adventure/zomes/integrity/dino_adventure/src/lib.rs) --- a scaffolded app that's been upgraded to 0.7 --- and adapt its dispatcher to your entry and link types.
```
ThetaSinner: Best not to reference dino-adventure here. It's not a canonical app and it is really just for testing.

I'd suggest instead that we recommend people to scaffold a new app if they want to see a clean copy of the target state.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+
+Now run:
+
+```bash
+npm run build:zomes
+```
+
+to see if all your updated dependencies and zome code compile.
+
+Sometimes dependency updates break builds! Cargo updates dependencies to the most recent versions that are _claimed_ to be compatible. However, sometimes package maintainers accidentally introduce a breaking change without labelling it as such. If you encounter build errors, this may be the cause. If this happens, try locking the incompatible dependency to a prior version in your `Cargo.toml` file, running `cargo update` again, and see if that fixes it.
```
ThetaSinner: That's not expected once we do a full release. I'd leave this bit out.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,700 @@
...
+-  webrtc_config: ~
++  request_timeout_s: 60
+   target_arc_factor: 1
+   report: none
+   advanced: ~
+-request_timeout_s: 60
+-chc_url: ~
+-db_sync_strategy: Resilient
++db_sync_level: Normal
++restore_chain_quorum: 3
```
ThetaSinner: This has a default value that probably shouldn't be overridden. Let's leave this out.

src/pages/resources/compatibility/holochain-0.7.md
ThetaSinner: Not waiting on anything — I had this wrong. `main` is updated: `package.json` there has `@holochain/client` `^0.21.0-rc.1` and `@holochain/hc-spin-rust-utils` `0.700.0-rc.1`, from commit `86ce025` ("feat: Update to 0.7.0-rc.3", 23 Jul).

Changed the row to link that commit, matching how the 0.6 table pins `4b2c383` rather than naming a branch. Fixed in f0c8f41.

src/pages/resources/upgrade/upgrade-holochain-0.5.md
ThetaSinner: You're right, and the tags do still exist — just not on the fork. `holochain-open-dev/tryorama` only has `v0.19.2`; `holochain/tryorama` still has the full tag history including `v0.18.1` and `v0.18.3`.

Pointed both links in this file at `holochain/tryorama` at `v0.18.3`, which is the version the [0.5 compatibility table](/resources/compatibility/holochain-0.5/) names as latest-compatible, and all three doc pages resolve there.

Note I also changed the `addPlayerWithApp`/`AppWithOptions` links on line 332, which were already pointing at `holochain-open-dev` `main` before this PR — same problem, and `v0.18.1` predates `AppWithOptions` so `v0.18.3` is the one that has all three. Fixed in b6000b9.

Left `build/testing-with-tryorama.md` on `holochain-open-dev` `main`, since that page isn't pinned to a Holochain version and its sibling links already point there.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: It is released — `EntryCreationData` is in rc.4 (`crates/hdi/src/flat_op/typed_action.rs`, with the `TryFrom<Action> for TypedAction<EntryCreationData>` impl), so no TODO needed.

The rename table further down already mapped `EntryCreationAction` to `TypedAction<EntryCreationData>`, but this intro bullet just said it was "gone", which reads like a removal rather than a replacement. Reworded to name the replacement. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Correct, and my row was wrong in both directions. At rc.4 `holo_hash` has no `sqlite`, `sqlite-encrypted` or `rusqlite` feature at all — the feature list is `default`, `full`, `test_utils`, `fixturators`, `fuzzing`, `hashing`, `kitsune2`, `serialization`, `encoding`, `holochain-wasmer`, `schema`. At 0.6.1 it had `sqlite = ["rusqlite", "rusqlite/bundled"]` and `sqlite-encrypted`, and `full` included `sqlite`. So this was a removal, not the implicit-feature rename I filed it as, and pointing people at `sqlite` was actively wrong.

`holochain_zome_types` is the same story — `sqlite`, `sqlite-encrypted` and `num_enum` are all gone at rc.4.

Dropped the row and replaced it with a sentence covering both crates. Fixed in 70f7199.

Since one row was wrong I checked the rest against `holochain-0.6.1` and `holochain-0.7.0-rc.4`; the other seven hold up:

| Row | 0.6.1 | rc.4 |
|---|---|---|
| `holo_hash` serde/serde_bytes | `serialization = ["holochain_serialized_bytes", "serde", "serde_bytes"]` | `dep:`-prefixed |
| `hdi` tracing/tracing-core | `trace = ["tracing", "tracing-core", …]` | `dep:`-prefixed |
| `holochain_integrity_types` subtle-encoding | `full = [… "subtle-encoding" …]` | `dep:subtle-encoding` |
| `holochain_zome_types` serde_yaml | `properties = ["serde_yaml"]` | `properties = ["dep:yaml_serde"]` |
| `holochain_trace` tokio/shrinkwraprs | `channels = ["tokio", "shrinkwraprs"]` | `dep:`-prefixed |
| `holochain_util` tokio | `default = [… "tokio" …]`, `time = ["tokio"]` | `tokio_helper = ["dep:tokio"]`, `time = []` |
| `holochain_nonce` subtle-encoding | `subtle-encoding = []` | `full = []` |

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Agreed, that framing is much easier to hold in your head than a rename table. Added after the `ActionHeader` paragraph:

> The `...Data` structs carry the same fields their 0.6 counterparts did, minus the four header fields that were lifted out of them. So when you're looking for a field, it either moved to `header` or kept its name on the data struct. Only a handful were renamed, and those are listed below.

with the "below" linking to the removed-field table in the `FlatOp` sub-types section. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Sloppy shorthand on my part — they were standing in for "and the rest of the variants", which is not something a reader can act on.

Removed both. The lead-in now states the pattern once ("the enum is now `ActionData`, matched on `action.data`, and each variant's payload struct gains a `Data` suffix") and the table spells out every variant, so `Dna`, `AgentValidationPkg`, `InitZomesComplete`, `OpenChain` and `CloseChain` are all named against their `...Data` counterparts. Checked against the `ActionData` definition in rc.4 so the list is complete. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: It does — `impl<D> TypedAction<D>` has `author()`, `timestamp()`, `action_seq()` and `prev_action()`, and `impl Action` has the same four. Both in rc.4.

Changed the diff to `action.author() != record.action().author()` (no `&`, since both sides are already `&AgentPubKey`) and reworded the lead-in to point at the accessors rather than telling people to reach into `header`. Also updated the rename table row from `action.header.author` to `action.author()`. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Confirmed — 3c39e86 (#5910) is in `develop` but not in `holochain-0.7.0-rc.4`, so it lands after what this guide documents. Left a `TODO(upgrade)` here naming the three specific things it changes, so this does not get missed at GA:

- restores `agent`/`new_key`/`original_key` as plain fields on the `CreateAgent`/`UpdateAgent` variants and drops the `Option`-returning methods this paragraph currently describes
- adds `Deref<Target = D>` to `TypedAction`, so the `action.data.x` reads in the table just above become `action.x`
- adds `TryFrom<Action>` and `try_from_action` for each single-variant data type, which supersedes the narrowing snippet further down

Kept the text describing rc.4, since that is what the version numbers in this guide point at. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Agreed. Dropped the dino-adventure link entirely.

The section now says to scaffold a throwaway app with the same entry and link types and copy its dispatcher across, with an RC admonition noting you will have to hand-edit for now — holochain/scaffolding#576 is still open and the newest release is v0.603.0, so there is no 0.7 scaffolding release yet. The existing `TODO(upgrade)` is now just "remove this note once a 0.7 scaffolding release exists".

The two dispatcher patterns below it are unchanged and stand on their own, so the section no longer depends on an external reference. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Removed. Fixed in 70f7199.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Removed from the config diff and dropped its bullet from the list below. Fixed in 70f7199.

`wasm_backend` is still described, since that one genuinely is a choice worth making, but it is documented as optional with "leave it unset" as the default advice.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,714 @@
...
+    ActionData::Create(create) => {
+        // `action.header.author`, `action.header.timestamp`, and `create.entry_type`
+    }
+    // ...
+}
+```
+
+`ActionHeader` holds `author`, `timestamp`, `action_seq` and `prev_action`. Everything else lives in the `ActionData` variant.
+
+The `...Data` structs carry the same fields their 0.6 counterparts did, minus the four header fields that were lifted out of them. So when you're looking for a field, it either moved to `header` or kept its name on the data struct. Only a handful were renamed, and those are listed [below](#flat-op-sub-types-carry-a-typed-action).
```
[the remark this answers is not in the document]
ThetaSinner: Not actioned — the fragment does resolve. This is MD051 applying GitHub's slug algorithm, which is not the one this site uses.

11ty generates `id="flat-op-sub-types-carry-a-typed-action"` for that heading, splitting on the case changes in `FlatOp` and `TypedAction`. Verified in the built output:

```
$ grep -c 'id="flat-op-sub-types-carry-a-typed-action"' _site/resources/upgrade/upgrade-holochain-0.7/index.html
1
```

The `hyperlink` fragment check that Netlify runs over the built HTML passes on this page (314 assertions, 0 failures), and that is the check that gates the deploy. MD051 reads the Markdown source and guesses the slug, so it disagrees with the actual renderer.

Worth noting the link was genuinely broken when I first wrote it — I had `#flatop-sub-types-carry-a-typed-action`, which `hyperlink` caught and I fixed in 70f7199 before this review ran.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,714 @@
...
+
+### Update the `validate` callback body
+
+The body of `validate` is scaffolded code that dispatches to the functions above, so most of this work is mechanical. Rather than rewriting it by hand, scaffold a throwaway app with the same entry and link types, and copy its dispatcher across.
+
+<!-- TODO(upgrade): remove this note once a 0.7 scaffolding release exists -->
+
+!!! info While 0.7 is a release candidate
+
+There's no 0.7 release of the scaffolding tool yet, so you'll have to hand-edit the dispatcher for now.
```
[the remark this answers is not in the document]
ThetaSinner: Fair — it did read as "do X" immediately followed by "you cannot do X". Reworded in 4a2b858 so the two parts fit together:

> The quickest way through it is to scaffold a throwaway app with the same entry and link types and copy its dispatcher across.
>
> !!! info While 0.7 is a release candidate
> The scaffolding tool has no 0.7 release yet, so that shortcut isn't available. Until it is, hand-edit your existing dispatcher using the two patterns below.

I kept the main sentence unconditional rather than taking the suggested "Once a 0.7 scaffolding release exists, scaffold..." phrasing, for two reasons: scaffolding is the recommendation the maintainer asked for in [this thread](https://github.com/holochain/docs-pages/pull/647#discussion_r3666483397), and the sentence then stays correct at GA when the admonition is deleted, rather than needing an edit of its own. That also matches how the Holonix ref is handled earlier in the guide — durable instruction in the body, RC caveat in an admonition with a `TODO(upgrade)`.

ThetaSinner:
Replying here to the outside-diff comment on `upgrade-holochain-0.7.md` (167-185), "Run `npm install` from the directory containing `ui/package.json`".

**Not actioned** — the premise is a misread. "Your project's root folder" means the reader's hApp, not this docs repo, and a scaffolded hApp declares `ui` as an npm workspace. From dino-adventure's root `package.json`:

```json
{
  "name": "dino-adventure-dev",
  "private": true,
  "workspaces": [
    "ui"
  ],
```

So `npm install` at the hApp root is the correct command: it resolves the `ui` workspace and updates the single root lockfile. There is no `ui/package-lock.json` to miss — workspaces deliberately hoist to one lockfile. `cd ui && npm install` would be the wrong advice, and the scripts in that same file (`npm run start --workspace ui`, `npm run package --workspace ui`) depend on the workspace wiring being installed from the root.

This also matches the [0.6 guide](/resources/upgrade/upgrade-holochain-0.6/), which gives the same instruction in the same place, so changing it here would make the two guides disagree.

src/pages/resources/howtos/running-network-infrastructure.md
```diff
@@ -129,28 +134,28 @@ If you use the same server for production and testing, you might end up writing
...
 +  bootstrapUrl: 'https://bootstrap.example.org/',
-   signalUrl: 'wss://dev-test-bootstrap2.holochain.org/',
--  relayUrl: 'https://iroh-relay-hc.holochain.org',
 +  relayUrl: 'https://bootstrap.example.org/',
-   iceUrls: ['stun:stun.l.google.com:19302','stun:stun.cloudflare.com:3478'],
    // ...
  });
 ```
 
+If you're coming from a 0.6 Kangaroo config, note that `signalUrl` and `iceUrls` are gone --- they configured the tx5/WebRTC transport, which [0.7 replaced with iroh](/resources/upgrade/upgrade-holochain-0.7/#conductor-config-file-changes).
```
ThetaSinner: Not exactly - Iroh was available in 0.6 but both Iroh and tx5 were options - in 0.7, we've removed tx5 and so its configuration is no longer accepted.

ThetaSinner: Corrected in 8d12b07. The sentence now reads:

> They configured the tx5/WebRTC transport, which 0.6 offered alongside iroh; [0.7 removes tx5](/resources/upgrade/upgrade-holochain-0.7/#conductor-config-file-changes), so its configuration is no longer accepted.

I checked the rest of the branch for the same error and it doesn't recur — the 0.7 upgrade guide already says "Iroh, over QUIC, is *now the only* network transport" and explains that `transport-iroh` "no longer exists because iroh is now compiled in unconditionally", both of which correctly imply iroh predates 0.7. The one loose phrase left is "reflecting the move to iroh relays" under `signalingServerUrl` renamed; I left it because it describes the rename's motivation rather than claiming iroh is new, but say the word if you'd rather it was explicit too.

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,683 @@
...
+Update the `hdk` and `hdi` version strings in the project's root `Cargo.toml` file:
+
+<!-- TODO(upgrade): Update these version numbers for new 0.7.x releases -->
+
+```diff:toml
+ [workspace.dependencies]
+-hdi = "=0.7.1"
+-hdk = "=0.6.1"
++hdi = "=0.8.0"
++hdk = "=0.7.0"
```
jost-s, line 107: Instead of pinning exact versions and pointing people to look up the current one, couldn't it be `hdi = "0.8"`?

src/pages/resources/upgrade/upgrade-holochain-0.7.md
```diff
@@ -0,0 +1,683 @@
...
++  advanced:
++    irohTransport:
++      relayAllowPlainText: true
+```
+
+## Subtle changes
+
+The following changes may not break your build, but they may require you to reassess whether your hApp will work as expected:
+
+* **Existing installs won't work.** The way zome definitions are serialized has changed, so the `DnaHash` of an otherwise-identical DNA is different in 0.7. Holochain's databases have also been renamed, with no migration path. Existing installs need their data cleared, and agents on 0.7 form a new network separate from the 0.6 one.
```
jost-s: Even though it's your expectation, migration hasn't been a thing and IMO doesn't deserve special mention, or rather it's been said in the introduction already.
```suggestion
```

jost-s (approved):
Soft comments, looking good

src/pages/resources/upgrade/upgrade-holochain-0.7.md
ThetaSinner: Possibly, since the WASM compile isn't deterministic anyway but I think that should be its own change and not rolled in here when the scaffolding generates the old code like this

